PR #21, opened 2015-12-31
Rotate cli command

saldoukhov:
Exposing rotate functionality previously available only from keeper
shell as a cli command. Example: keeper rotate —match=“MySQL Dev”


keepercommander/__init__.py
```diff
@@ -14,7 +14,12 @@
 @click.version_option(version=__version__)
 @click.pass_context
 def main(ctx, debug, server, user, password, config):
-
+    '''
+    \b
```
saldoukhov, line 18: This keeps comment as block

PR #28, opened 2016-11-11
record print and bug fixes

saldoukhov:
The following changes and fixes:
- equals sign is stripped after b64 encode
- get-rec command is added
- rotate command does not require  plug anymore
- —print option is added to rotate command


kelygirl:
Ray what are you doing? All I still want is your time. Guess its never
gonna be anything you give to me....

PR #113, opened 2019-02-28
Updated class RecordUploadAttachmentCommand

tonylanglet:
There is an issue when trying to add a file as an attachment. When using the argument --file the current setup reads through the string character by character and tries to add a file with the name of each character. By removing the loop through and adding the kwargs['file'] variable you'd be able to upload one file. A comma seperated array could probably be a solution for this class if want to add more than one file at the time. 

Current result when trying to add a file from a regular windows file path like <c:\users\tony\desktop\pythonversion.png>:
    + CategoryInfo          : NotSpecified: (File C does not exists:String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError
 
File : does not exists
File \ does not exists
File u does not exists
File s does not exists
File e does not exists
File r does not exists
File s does not exists
File \ does not exists
File t does not exists
File o does not exists
File n does not exists
File y does not exists
File \ does not exists
File d does not exists
File e does not exists
File s does not exists
File k does not exists
File t does not exists
File o does not exists
File p does not exists
File \ does not exists
File p does not exists
File y does not exists
File t does not exists
File h does not exists
File o does not exists
File n does not exists
File v does not exists
File e does not exists
File r does not exists
File s does not exists
File i does not exists
File o does not exists
File n does not exists
File . does not exists
File P does not exists
File N does not exists
File G does not exists
No files to upload

### DIFF
--- a/keepercommander/commands/record.py
+++ b/keepercommander/commands/record.py
@@ -668,12 +668,11 @@ class RecordUploadAttachmentCommand(Command):
 
         files = []
         if 'file' in kwargs:
-            for name in kwargs['file']:
-                file_name = os.path.abspath(os.path.expanduser(name))
-                if os.path.isfile(file_name):
-                    files.append(file_name)
-                else:
-                    api.print_error('File {0} does not exists'.format(name))
+            file_name = os.path.abspath(os.path.expanduser(kwargs['file']))
+            if os.path.isfile(file_name):
+                files.append(file_name)
+            else:
+                api.print_error('File {0} does not exists'.format(kwargs['file']))
         if len(files) == 0:
             api.print_error('No files to upload')
             return


keepercommander/commands/record.py
```diff
@@ -668,12 +668,11 @@ def execute(self, params, **kwargs):
 
         files = []
         if 'file' in kwargs:
-            for name in kwargs['file']:
```
sk-keeper, line 671: 'file' parameter is defined as
`upload_parser.add_argument('--file', dest='file', action='append', required=True, help='file name to upload.')`
So, file parameter is expected to be a list of strings. 
Probably, 'argparse' package for some Python versions on Windows do not parse a single 'file' parameter as a list.
The proper solution would be checking type of kwargs['file'] and run the old code if it's a list or new code if it's a str

sk-keeper: What Python version do you use?

tonylanglet: Makes sense...
I'm uisng the python version which is included in WinPy-3670 which is python-3.6.7.amd64

sk-keeper: I cannot duplicate this issue in WinPy-3670.
`upload-attachment --file="C:\\Users\sk\aa" NmMRFMQOU5tHJMQXtMQRYL`
command successfully uploads file attachment.
Is it possible that system 'argparse.py' script is altered or conflicting package is installed with pip?
It would be nice to know the problem before applying the patch.

tonylanglet: @sk-keeper Thanks for taking your time to investigate this so called problem I encountered. I've now found what the issue was from my part. As I'm creating a wrapper for powershell there was one of the jumps between systems that made the input wrong. As you mentioned in the first part dest='file' was what was lacking in my script and therefore it failed. 

You could remove this pull-request as it is not needed if you do it the right way.

PR #120, opened 2019-03-28
Audit log export to Azure Log Analytics

LukasNespor:


keepercommander/commands/enterprise.py
```diff
@@ -120,7 +122,7 @@ def register_command_info(aliases, command_info):
 
 
 audit_log_parser = argparse.ArgumentParser(prog='audit-log', description='Export enterprise audit log')
-audit_log_parser.add_argument('--target', dest='target', choices=['splunk', 'syslog', 'syslog-port', 'sumo'], required=True, action='store', help='export target')
+audit_log_parser.add_argument('--target', dest='target', choices=['splunk', 'syslog', 'syslog-port', 'sumo', 'log-analytics'], required=True, action='store', help='export target')
```
sk-keeper, line 125: Would you mind 'log-analytics' to be renamed to 'azure'?

LukasNespor: Well Azure is cloud platform and Log Analytics is a service, so it would be confusing for users familiar with Azure and Log Analytics. Also other services in Azure could be used for storing logs. Maybe 'azure-la'?

sk-keeper: 'azure-la' sounds good. 'log-analytics' may be confusing for users unfamiliar with Microsoft's Azure stack

PR #121, opened 2019-03-28
Audit log export to JSON format

LukasNespor:


keepercommander/commands/enterprise.py
```diff
@@ -1335,6 +1335,42 @@ def chunk_size(self):
...
+    def export_events(self, props, events):
+        filename = props['filename']
+        logf = open(filename, mode='w')
+
+        try:
+            json.dump(events, logf)
+        finally:
+            logf.flush()
+            logf.close()
+
```
sk-keeper, line 1379: export_events overrides already exported events. 
The export file contains events from the last chunk only.
Chunk is 1000 events.

LukasNespor: Yes, that is true. This behavior is also stated in readme. We have use case in our company, that we need only thouse incremental logs and process them with different application before sending to some system. It is much easier to deserialize JSON then parsing lets say for example syslog format. So I thought, that it may be helpful to someone else.

LukasNespor:
Sorry, I mistakenly deleted the branch :) but restored it immediately back.

keepercommander/commands/enterprise.py
sk-keeper: Even incremental sync is going to be truncated to the last chunk if number of audit events exceeds 1000. The result JSON file will never contain more than 1000 events.
For instance: there are 5100 audit event to be downloaded. The resulting export file is going contain last 100 events. The previous 5000 events will be lost

sk-keeper: Probably, exporting audit log to JSON requires output file rotation similar to the one used in export to syslog

LukasNespor: Oh, I didn't realize that it is exporting events in chunks and that file is overwritten with every iteration. See latest commit.

LukasNespor:
@sk-keeper Is there any issue with the code that I can fix?

sk-keeper:
> @sk-keeper Is there any issue with the code that I can fix?

I just back from vacation. This pull request has merge conflicts. No problem, I'll fix it

PR #241, opened 2020-05-08
Send yubikey prompts to the tty, as with user/pass

### LINKED ISSUE #238: U2F fails in batch mode
When using batch mode with a 2FA Yubikey enabled account, keeper prompts for the U2F touch however immediately falls through to a 2FA code prompt, as if "Enter" had been pressed.

env: 
```
% sw_vers 
ProductName:	Mac OS X
ProductVersion:	10.15.4
BuildVersion:	19E287
% keeper --version                                         
Keeper Commander, version 4.31
```

- keeper installed via `pip3`, per readme

repro:
```
% rm config.json
% echo 'whoami' | keeper --user 'user@example.com' -
Password:

Touch the flashing U2F device to authenticate or press Enter to resume with the primary two factor authentication...
Two-Factor Code:
```

- same thing when using `--batch-mode shell`
- fwiw, the yubikey begins flashing at the expected time; entering a valid code succeeds

ad8-bdl:


ad8-bdl:
Fixes #238

sk-keeper:
This change is going to fail on Windows. `/dev/tty`
`print('OK')` will be replaced with with `logging.info('OK')`
You are right there is no need to print it to stdout

sk-keeper:
I cannot merge this pull request as is. It contains two commits. Can you squash your commits into one and make commit description correct?

ad8-bdl:
Done, ta.

sk-keeper:
Merged into development repository

ad8-bdl:
This appears to have not actually been merged?

PR #262, opened 2020-06-24
Support for ssh rotation in Windows

shreyankkr:
This will enable ssh plugin to support rotations possible when commander is running in Windows OS. This uses paramiko module (http://docs.paramiko.org/en/stable/) to establish ssh connection, which uses socket based communication. This also uses a simple thread to enable reading from shell and writing back on passwd prompt.

### DIFF
--- a/keepercommander/plugins/ssh/README.md
+++ b/keepercommander/plugins/ssh/README.md
@@ -6,9 +6,13 @@ The SSH plugin for Keeper Commander gives you the ability to rotate any local or
 ### Dependencies 
 
 1) Install the below modules
-
+If Keeper Commander running in Unix based OS
 ```
 pip3 install pexpect
+```
+For Windows OS
+```
+pip3 install paramiko
 ```
 
 2) Add the following Custom Fields to the record that you want to rotate within Keeper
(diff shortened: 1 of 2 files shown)


b1tninja:
Love it.  A minor security enhancement, the AutoAcceptPolicy() I think we would need to change, and I worry about the rotation code typing out the password to the ssh endpoint, but will talk to the team about merging this with sufficient "WARNING" notices to the user, since this is the more portable solution... absent an endpoint or "agent". Additional thoughts I had when reviewing, this is obviously catered toward *nix, there is a more rare Windows SSH server.

Thanks, I'll talk with the team about if we can merge this or not, I'd like to as a opt-dep sort of plugin

ESA-Security-Group (approved):
#332  ⚜️🪄✨🧛🏼‍♀️👩‍❤️‍💋‍👩👭🧍🏼‍♀️🧚🏼‍♀️🧜🏼‍♀️🧝🏻‍♀️👸🏼👸🏻🧑🏽‍🎤👩🏽‍🦰👩🏼🧍🏽‍♀️Give Ambjörn🥇🧛🏼🩸Sköld pussy!💦🌹✨

ESA-Security-Group (approved):
EDITED

kihneman:
Closing in favor of https://github.com/Keeper-Security/Commander/pull/739

PR #281, opened 2020-09-04
Ability for MSP Admin login or execute commands as MC

maksimu:


keepercommander/commands/enterprise.py
```diff
@@ -242,7 +242,7 @@ def __init__(self):
         self.team_keys = {}
 
     def execute_args(self, params, args, **kwargs):
-        if params.enterprise:
+        if params.enterprise or params.enterprise_id > 0:
```
sk-keeper: It looks like this change is not needed. If params.enterprise_id > 0 then params.enterprise should always be populated.

keepercommander/cli.py
```diff
@@ -70,7 +78,86 @@ def display_command_help(show_enterprise = False, show_shell = False):
...
+            mc_id = int(mc_id.split()[1])                      # get just the ID
+        except AttributeError:
+            logging.error("No Managed company provided")  # apply your error handling
+            raise
+
+        cur_msp_params = params if msp_params is None else params
+
+        query_enterprise(cur_msp_params)
+        query_msp(cur_msp_params)
+
```
sk-keeper: "query_enterprise" may take awhile to complete. 
Is it really needed in this context?
We are switching to MC and we are going to load MC data anyway.
We need to reload enterprise data when we back to MSP, though.

maksimu: This was done to make sure data is refreshed even when switching. Removed these lines.
Also added reload of the data when switching back to MSP

keepercommander/commands/enterprise.py
maksimu: This check was done before synching of the data was added to another method. Removed.

PR #284, opened 2020-09-07
Admin device approval

sk-keeper:


keepercommander/params.py
```diff
@@ -97,7 +97,6 @@ def __init__(self, config_filename='', config=None, server='https://keepersecuri
         self.enterprise = None
         self.enterprise_id = 0
         self.msp_tree_key = None
-        self.msp = None
         self.prepare_commands = False
```
sk-keeper, line 100: It looks like msp is not used in the code

keepercommander/api.py
```diff
@@ -1691,7 +1692,7 @@ def fix_data(d):
 
     request = {
         'command': 'get_enterprise_data',
-        'include': ['managed_companies', 'licenses', 'nodes']
+        'include': ['managed_companies']
     }
```
sk-keeper, line 1696: MSP is the enterprise that supports an additional data type: managed_companies
Do we really need to put it to the separate method?

PR #291, opened 2020-09-29
Allow specifying path to config file as environment variable KEEPER_CONFIG_FILE

armarti:
I'd like to put something like in my shell's rc file, or maybe at the root of a deep git repo:

```
export KEEPER_CONFIG_FILE="$one_central_safe_place/.keeperrc.json"
```

Then I can run `keeper` anywhere in the filesystem and it'll always use that `.keeperrc.json` for reading / caching credentials.

I could see this as dangerous for a novice who might not realize they have very important credentials stored perhaps unprotected in clear text, but this tool seems to be targeted at tech professionals who ought to know how to take measured risks and what's safe and what's not.

I'll admit though I'm only barely familiar with KeeperCommander's API, so perhaps there's a security issue here I'm overlooking? Or is this a safe convenience to add?

Thanks,
DCM

### DIFF
--- a/README.md
+++ b/README.md
@@ -1197,7 +1197,7 @@ $ keeper --config=my_config_file.json
 
 ### Advanced Configuration File
 
-By default, Keeper will look for a file called ```config.json``` in the current working directory and it will use this file for reading and writing session parameters. For example, if you login with two factor authentication, the device token is written to this file. The configuration file loaded can also be customized through the ```config``` parameter. The config file can also be used to automate and schedule commands.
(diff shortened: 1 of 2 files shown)


armarti:
eh?

b1tninja:
Hey, I just joined this project and am reviewing your changes. First thanks, and apologies for the delay. I agree there should be a way to configure the location of the config file. I like the idea of using the environment variable, it's cross-platform, doesn't demand any OS specific knowledge like system root drive, and isn't the CWD. I am in favor of merging your commit. A command line option like --config. It does have some security and user implications around config file precedence ie. "which config file gets loaded", so I'll review with team, but my thinking is if the user has setup their environment that way, it's intentional... I don't see much potential for abuse so long as we don't allow reconfiguration of the endpoints or anything like that in the config at any point in the future-- and I don't think we would, but again, thanks I'll review again with team before merging

b1tninja (approved):
I'll see about giving config_filename a kwarg with a default None value

PR #294, opened 2020-10-15
bug fix by selecting correct salt for SSO login w/ master password

maksimu:


keepercommander/api.py
```diff
@@ -1813,3 +1813,11 @@ def login_and_get_mc_params(params, mc_id):
     query_enterprise(mc_params)
 
     return mc_params
+
+
+def get_correct_salt(salts):
+    if len(salts) > 1:
+        salt = next((s for s in salts if s.name.lower() == 'alternate'), None)
```
sk-keeper: it means 'crash if there is no salt named alternate' 
Ideally we should try all salts
otherwise the following would be better
`def get_correct_salt(salts): 
  return next((s for s in salts if s.name.lower() == 'alternate'), salts[0])`

maksimu:
resolved conflict in another pull request

PR #318, opened 2021-01-05
Send yubikey prompts to the tty, as with user/pass; "OK" to logging.info

ad8-bdl:


b1tninja (approved):
Thanks, this inspired me to look through the code and I noticed at least one other place we can be using getpass, so I'll be fixing that soon.

PR #328, opened 2021-02-21
Fixing 2fa prompt

maksimu:


b1tninja (commented):
I already have a pr for this. I'd like to revert this in favor of https://github.com/Keeper-Security/Commander/pull/324

PR #330, opened 2021-02-24
Error while adding a user to administrator group

sk-keeper:
To reproduce the issue the administrator group should have a role key assigned to the role (it means Account Transfer or Manage Companies privilege)
It needs to be an old `role_key` not new `role_key2` 

Also, it seems that the Commander does not support `role_key2`.


maksimu:
Replacing w/ #331

PR #347, opened 2021-04-14
fixed provision_user_in_enterprise

idimov-keeper:


maksimu:
Tested. Works as expected

PR #360, opened 2021-04-30
Improved help for --custom option

idimov-keeper:


keepercommander/commands/record.py
```diff
@@ -104,7 +104,7 @@ def register_command_info(aliases, command_info):
 add_parser.add_argument('--pass', dest='password', action='store', help='password')
 add_parser.add_argument('--url', dest='url', action='store', help='url')
 add_parser.add_argument('--notes', dest='notes', action='store', help='notes')
-add_parser.add_argument('--custom', dest='custom', action='store', help='custom fields. name:value pairs separated by comma. Example: "name1: value1, name2: value2"')
+add_parser.add_argument('--custom', dest='custom', action='store', help='add custom fields. JSON or name:value pairs separated by comma. CSV Example: --custom "name1: value1, name2: value2". JSON Example: --custom \'{"name1":"value1", "name2":"value2,3,4"}\'')
```
maksimu: how about changing `"value2,3,4"` to `"value:2,3,4"?`

maksimu: That way we can show that it is possible to add values that can contain colons and commas

PR #373, opened 2021-06-02
Update README.md

ZaneBond:
Updating install instructions.


jsupun (commented):
The only question I have is does pip comes with the Windows python install? Sometime python installed don't install pip and have to "bootstrap" it in ( https://bootstrap.pypa.io/get-pip.py).

maksimu:
made a new pull request with more details

https://github.com/Keeper-Security/Commander/pull/374/files

PR #383, opened 2021-07-08
Improve password generation

maksimu:
Improve random password generation by using builtin python function for python 3.6+ and use legacy one for python any other python versions

### DIFF
--- a/keepercommander/generator.py
+++ b/keepercommander/generator.py
@@ -7,7 +7,7 @@
 # Keeper Commander 
 # Contact: ops@keepersecurity.com
 #
-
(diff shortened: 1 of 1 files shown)


dstromberg (approved):
Looks good.  We already talked about the default value of sampleString tracebacking.  I didn't mention previously: for long inputs a += b is quadratic for immutable types, but that may or may not matter here.  I usually replace it with list_ = []; for i in whatever: list_.append(b); a = ''.join(list_).  This only matters if sampleLength is large, and only some python implementations have the problem, not all.

PR #405, opened 2021-08-06
Added --update option to import command.

dstromberg:
go rule becomes report rule. pylint exit status is ignored

    Parse and pass around update/update_flag

    Many small linter fixes. Attempts to update records through older/import_folders_and_records and fails.  Added result status output

    Introduced PYTHON_FILES variable, rather than computing it in multiple rules

    A little refactoring to simplify matters

    import appears to work, but no --update at all.

    import --update works for a single update record

    Misc.

    Split some code out into functions.
    Decoupled import-or-update-control from the import-and-update.
    Works on import and at least one update.

    Eliminated most FIXMEs

    Improved error reporting a bit. Replaced a manual hashlib use with build_record_hash

    Added 2 FIXME's and a comment about an assert

    Added authentication data to execute_update_record. Improved some variable names

    Ignore no-member from keepercommander.folder_pb2

    Encrypt correctly, I think.  Need to preserve notes and custom fields though.

    Fixed a type comment that was perplexing pylint.  Removed the final newline

    Added targeted-report rule and some comments

    Preserve notes and custom

    Removed pudb.set_trace()

    Removed 2 pprint invocations

    Changed a comment to a FIXME

    Introduced a LOCALE global

    Downcase statuses. Use api.LOCALE. Small change to notes and custom preservation

    Eliminated pudb.set_trace()

    Deleted an internal-only URL

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,12 +1,22 @@
-go2: tags
-	python3 keeper.py
+PYTHON_FILES := ${shell find . -name '*.py' -print}
 
-tags:
-	ctags $$(find * -name '*.py' -print)
+go2: tags targeted-report
+	python3 -m pudb keeper.py
+	# python3 keeper.py
 
-go:
+tags: ${PYTHON_FILES}
+	# Create the tags file, for the benefit of vi/vim/neovim.
+	ctags ${PYTHON_FILES}
+
+.PHONY: targeted-report
+targeted-report:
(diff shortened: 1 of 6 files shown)


sk-keeper (commented):
nm

sk-keeper (approved):

PR #419, opened 2021-08-11
KC-194: Automatically log back in if login times out, but only if "relogin": true is in config.json.

dstromberg:
KC was mistakenly trying to treat v3 enforcements like v2 enforcements

Much linter-response. Did not do the docstrings though

Added a v2 login to the go2 rule

Try to get relogin from params.config

On successful login, set params.login_time

Added possible_relogin function and use of same. Do not set tmer when relogin. Elimninated an unused variable

Set params.login_time on successful login

Added self.login_time, which defaults to None

Ignore tags

When setting a timeout, stash it in params. Some long lines reformatted to < 132 columns. logout_timer value obtained locally, not from the server; the server wanted it to be 120 unconditionally. Eliminated an unused variable. Added a FIXME about send/write return values


sk-keeper:
KC-194 Jira title is "Add `keep-alive` command to extend session where user is currently logged in"
The code changes do not contain any command.

dstromberg:
Sergey: No one is doing v2 logins anymore, so we do not need to worry about those.
There is keepalive doc in the wiki
cli.py
Need to track server time, last time we accessed the server.
If it's been a long time, we need to issue a keepalive command.
Everything ends up in execute_rest()
JSON vs REST.
Reset the timer in communicate_rest() for REST and communicate() for JSON
If 5 minutes or 10 minutes remaining, do keepalive request in while True: command = '' loop.
tmer becomes irrelevant / redundant.
If server times out (almost), run keepalive request.
Don't worry about changes in duration during a session.

PR #428, opened 2021-08-17
KC-194-2-add-keep-alive-command: Added keepalive. Removed some unused variables

dstromberg:


keepercommander/api.py
```diff
@@ -1596,9 +1649,12 @@ def communicate_rest(params, request, endpoint):
     raise KeeperApiError('Error', endpoint)
 
 
-def communicate(params, request):
+def communicate(params, request, skip_ttk=False):
```
sk-keeper: What is the point of adding skip_ttk parameter?
It would be nice to keep the existing method signatures unchanged.
This method is one the most used methods in the app.

keepercommander/loginv3.py
```diff
@@ -882,12 +882,12 @@ def set_user_setting(params: KeeperParams, name: str, value: str):
...
 
     @staticmethod
-    def accountSummary(params: KeeperParams):
+    def accountSummary(params: KeeperParams, skip_ttk=False):
 
         rq = proto_as.AccountSummaryRequest()
         rq.summaryVersion = 1
 
-        rs = api.communicate_rest(params, rq, 'login/account_summary')
+        rs = api.communicate_rest(params, rq, 'login/account_summary', skip_ttk)
```
sk-keeper, line 890: ?

keepercommander/api.py
```diff
@@ -1571,13 +1572,65 @@ def prepare_record_v3(params, record):
...
+            self.server_logout_timer_window = get_server_logout_timer_window(params)
+            self.time_of_last_activity = current_time
+
+
+TTK = TimeToKeepalive()
+
+
+def get_acct_summary_dict(params):
+    """Get account summary data via protobufs."""
+    acct_summary = loginv3.LoginV3API.accountSummary(params, skip_ttk=True)
```
sk-keeper, line 1602: account_summary is pretty heavy API request

keepercommander/api.py
```diff
@@ -1571,13 +1572,65 @@ def prepare_record_v3(params, record):
...
+
+    def update(self, params):
+        """Update the timer, and possibly issue a keepalive."""
+        current_time = time.time()
+        if (
+            self.server_logout_timer_window is None or
+            (self.server_logout_timer_window / 3) + self.time_of_last_activity < current_time
+        ):
+            send_keepalive(params)
+            self.server_logout_timer_window = get_server_logout_timer_window(params)
```
sk-keeper, line 1591: account_summary is going to be called again and again

keepercommander/api.py
dstromberg: Agreed.  That's why it's called conditionally.  What would you do differently?

PR #441, opened 2021-08-19
Linting api.py

dstromberg:
apy.py now passes pylint and pyflakes.  It does not pass mypy, pycodestyle or pydocstyle.

### DIFF
--- a/keepercommander/api.py
+++ b/keepercommander/api.py
@@ -102,6 +102,7 @@ def login(params):
             logging.debug('No auth verifier, sending pre-auth request')
             try:
                 pre_login_rs = rest_api.pre_login(params.rest_context, params.user)
+                # pylint: disable=no-member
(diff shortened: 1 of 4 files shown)


keepercommander/api.py
```diff
@@ -1901,9 +1907,9 @@ def add_record_v3(params, record, **kwargs):
     if record_links:
         ra.record_links.extend(record_links)
     if audit:
-        #ra.audit = audit # Assignment not allowed to field "audit" in protocol message object.
+        # ra.audit = audit # Assignment not allowed to field "audit" in protocol message object.
         ra.audit.version = audit.version
-        ra.audit.data = audit.data
+        ra.audit.data = audit.data.encode('UTF-8')
```
sk-keeper, line 1912: It seems this line is going to crash.

PR #445, opened 2021-08-20
Change temporary file name created by connect command. KC-230

sk-keeper:


keepercommander/commands/utils.py
```diff
@@ -1746,8 +1746,9 @@ def get_parameter_value(params, record, parameter, temp_files, non_shared):
                 logging.error('Attachment file \"%s\" not found', file_name)
                 return None
             body = ConnectCommand.attachment_cache[file_name] # type: bytes
+            prefix = (kwargs.get('endpoint') or file_name) + '.'
             if parameter.startswith('file:'):
-                tf = tempfile.NamedTemporaryFile(delete=False)
+                tf = tempfile.NamedTemporaryFile(delete=False, prefix=prefix)
```
sk-keeper, line 1751: It seems that adding delete=False is enough to make temporary file visible on Windows.

kihneman: Yes. I did the same thing in PR to fix tests for Windows:
https://github.com/Keeper-Security/Commander/pull/435
For the tests, you will still want to clean up the temp file, thought.

kihneman (approved):

PR #450, opened 2021-08-25
Raise error for empty title when adding record (KC-211)

kihneman:


keepercommander/commands/record.py
```diff
@@ -209,6 +209,8 @@ def execute(self, params, **kwargs):
         if not force:
             if not title:
                 title = input('...' + 'Title: '.rjust(16))
+                if len(title) == 0:
```
sk-keeper: I don't think this change is enough.
The empty title can be created when --force parameter is passed.
I can see recordv3.py that has RecordAddCommand.

kihneman: There is a check for a recordv3 title here: https://github.com/Keeper-Security/Commander/blob/ff396298ad720006b4ba7a33b9b86549012552af/keepercommander/commands/recordv3.py#L375

keepercommander/commands/record.py
```diff
@@ -215,6 +215,9 @@ def execute(self, params, **kwargs):
                 password = input('...' + 'Password: '.rjust(16))
             if not url:
                 url = input('...' + 'Login URL: '.rjust(16))
+        if not title:
+            if len(title) == 0:
+                raise CommandError('add', 'Invalid title. Expected non-empty string.')
```
sk-keeper: My Vault> add -v2 -f                                                                                                                                                                                                                                  
An unexpected error occurred: object of type 'NoneType' has no len(). Toggle debug to print traceback

kihneman: I didn't mean to have the second conditional in there.

sk-keeper (approved):

PR #454, opened 2021-08-26
KC-196 allow cd command to navigate to folders with slash or quote in the title

dstromberg:


Makefile
```diff
@@ -1,7 +1,12 @@
 PYTHON_FILES := ${shell find . -name '*.py' -print}
 
 go2: tags targeted-report
-	python3 -m pudb keeper.py
+	# python3 -m pytest -s tests/test_vault_commands.py
```
kihneman: Can we just simply run `pytest` here on one line? Any test configuration to specify what tests are run can be specified in pytest.ini. I think that would make things much simpler.

unit-tests/test_subfolder.py
```diff
@@ -0,0 +1,283 @@
...
+        assert result == ['a/b/c/d']
+
+    def test_a_slash_slash_b_slash_slash_c_slash_slash_d_slash_e(self):
+        """Test a//b//c//d ."""
+        list_ = ['a', '', 'b', '', 'c', '', 'd', 'e']
+        result = subfolder.handle_subsequent_slash_slash(list_)
+        assert result == ['a/b/c/d', 'e']
+
+
+if __name__ == '__main__':
```
kihneman: Why is this necessary? It might be helpful to reduce unnecessary boilerplate code.

keepercommander/subfolder.py
```diff
@@ -32,55 +32,140 @@ def find_folders(params, record_uid):
                 yield fuid
 
 
+def handle_initial_slash(params, folder, components):
```
kihneman: Wouldn't Python re be simpler and easier to maintain here?
https://docs.python.org/3/library/re.html

Makefile
dstromberg: To the best of my knowledge, the Makefile is really only there because I use it a lot.  I have a vim macro that writes the current file, clears the screen, and runs the first rule in the Makefile.  As I'm working, I change that first rule a lot, to reflect what I want most convenient access to at the moment.  So running pytest in its entirety is good sometimes, but so is just running the tests I want for the current file I'm editing.

unit-tests/test_subfolder.py
dstromberg: Why is what necessary?  The if \_\_name\_\_ == '\_\_main\_\_'?

I put it there to facilitate running a debugger on a specific test.  ISTR adding a pudb.set_trace() does not play nicely with pytest, probably due to output capture.

Implication: I like pudb.

keepercommander/subfolder.py
dstromberg: I guess I'm missing how regexes would be simpler.  Are you suggesting that we try some re.sub's?  And hook that up to changing the folder when necessary with an re.match?  It seems like we're talking about 3 or 4 regexes to me.  I guess it might reduce the SLOC by a line or two - I'm not 100% sure it'd be smaller.

I confess: despite nonzero regex proficiency, I'm not the biggest regex fan.  I loved my Automata Theory classes though.

Aren't regexes often considered unpythonic?  https://www.reddit.com/r/learnpython/comments/96htn8/are_regexes_thhe_pythonic_way_to_manipulate/
They're undeniably useful in some cases, but not nearly as much as Perl would have people doing.

dstromberg: It seems like to convert to a path of folders and an as-yet-uncreated remainder of the path using regexes, we'd need to keep track of how far we've matched so far and search from there on out, in a loop. It just doesn't feel regexy to me.  It occurs to me that maybe we could create a list of folders indexed by their quoted names instead of their literal names....  Not sure about that.

keepercommander/autocomplete.py
```diff
@@ -233,8 +247,8 @@ def get_completions(self, document, complete_event):
                                         n = r.get('display_name') or ''
                                         if len(n) > 0:
                                             if n.lower().startswith(name) and len(name) < len(n):
-                                                if extra.get('escape_space'):
-                                                    n = n.replace(' ', '\\ ')
+                                                if extra.get('escapes'):
+                                                    n = escape_string(n)
                                                 d = n
```
sk-keeper, line 243: We need always escape quotes both single and double.
It is safe if quotes are always escaped. 
My Vault > cd "Fol&lt;Tab&gt; 
is going to create invalid path when folder name contains a double quote like `Folger"s Coffee`
Theoretically it would be possible to detect the if the first character is one of the quotes and escape only the other quote but escaping quotes always is fine too

sk-keeper: Forward slash is also not escaped when folder path starts with quote.

sk-keeper: Also may be related. Autocompletion does not go beyond the folder that contains quotes

dstromberg: Scope creep?  Maybe this should be a different Jira ticket.  Essentially the same problem is in master too.  It's not just cd that's showing the issue.

sk-keeper: I don't see any line of code in this PR directly related to the `cd` command.
What scope creep you are talking about?

dstromberg:
BTW, it's not perfect.  But it's passing all of the tests I've thrown at it, other than putting an apostrophe inside double quotes. fix_input still avoids that.

Makefile
kihneman: This shouldn't be in the repo if it's for your personal use. You can use a .gitignore.

unit-tests/test_autocomplete.py
```diff
@@ -0,0 +1,196 @@
...
+
+
+def check_roundtrip_wo_in_double_quotes(string):
+    """Test string for escape -> unescape, both without double quotes."""
+    escaped = autocomplete.escape_string(False, string)
+    unescaped = autocomplete.unescape_string(False, escaped)
+    assert string == unescaped
+
+
+class TestAutocompleteQuotingWithInDoubleQuotes(TestCase):
```
kihneman: Please use parameterized tests to simplify the following tests. See https://github.com/Keeper-Security/Commander/blob/master/unit-tests/commands/helpers/test_timeout.py

kihneman (commented):
I think there is still a number of issues with the PR. I am mentioning a few things that you could start with.

Makefile
dstromberg: Done.

unit-tests/test_autocomplete.py
dstromberg: Done.

unit-tests/test_subfolder.py
```diff
@@ -0,0 +1,202 @@
...
+    """Create a fake params instance for testing."""
+    params = Mock()
+    (params.folder_cache, root_bfn, cd_tests_bfn) = folder_cache()
+    params.current_folder = ''
+    params.root_folder = root_bfn
+
+    return params, root_bfn, cd_tests_bfn
+
+
+class TestSubfolderTryResolvePath(TestCase):
```
kihneman: These tests would also benefit from being parameterized.

dstromberg: Done.

dstromberg:
Is there anything else that needs to happen on this?  Kevin, you seemed to be saying you had a lot of issues with it.

kihneman:
> 
> 
> Is there anything else that needs to happen on this? Kevin, you seemed to be saying you had a lot of issues with it.

I can take some time to go over it again this weekend in light of our decision to ignore the ambiguity with using "//".

kihneman:
I did wonder why updates for autocompletion are included in this PR. Generally smaller and more concise PRs are better as opposed to addressing multiple issues in one PR.

dstromberg:
I tend to agree: this is like two separate issues in one Jira/PR.  But that's how Sergey wanted it, so that's what I did.

keepercommander/subfolder.py
```diff
@@ -32,55 +32,83 @@ def find_folders(params, record_uid):
...
-                break
-            if idx < 0:
-                path = ''
-                break
-
-            path = path[idx+1:]
-            start = 0
-
-    return folder, path
+    folder = (
```
kihneman, line 93: I think the following would be a bit simpler and you don't need the extra functions added to subfolder.py:
(import of posixpath is needed)
```
    no_slash_path = path.replace('//', '\0').strip()
    if no_slash_path.startswith('/'):
        abspath = no_slash_path[1:]
    elif params.current_folder == '':
        abspath = no_slash_path
    else:
        base_path = get_folder_path(params.current_folder)
        abspath = base_path + '/' + no_slash_path
    normpath = posixpath.normpath(abspath)

    if normpath.startswith('.'):
        return params.root_folder, ''
    else:
        subfoldernames = normpath.split('/')
        folder = params.root_folder
        for i, subfoldername in enumerate(subfoldernames):
            subfoldername = subfoldername.replace('\0', '/').strip()
            subfolder = [params.folder_cache[s] for s in folder.subfolders if params.folder_cache[s].name == subfoldername]
            if len(subfolder) == 0:
                break
            else:
                folder = subfolder[0]
        else:
            return folder, ''

        return folder, '/'.join(subfoldernames[i:]).replace('\0', '//')
```

dstromberg: The separate functions enable thorough testing and complexity reduction.  The suggested alternative is far from simpler; take pity on the code's cyclomatic complexity: https://en.wikipedia.org/wiki/Cyclomatic_complexity

kihneman: Well break the above into functions. I still think it's simpler.

sk-keeper (approved):

PR #456, opened 2021-08-26
If keepalive fails, output a warning

dstromberg:


keepercommander/ttk.py
```diff
@@ -42,7 +44,10 @@ def update(self, params):
                 self.server_logout_timer_window is not None and
                 (self.server_logout_timer_window / 2) + self.time_of_last_activity < current_time
         ):
-            api.send_keepalive(params)
+            try:
+                api.send_keepalive(params)
+            except keepercommander.error.KeeperApiError:
+                logging.warning('Error sending keepalive to server. You should exit and log back in.')
```
maksimu: Why not use message that is returned from the server?

PR #457, opened 2021-08-27
Prompt for password rotation with --match option unless --force option (KC-228)

kihneman:


keepercommander/utils.py
```diff
@@ -76,3 +76,16 @@ def create_auth_verifier(password, salt, iterations):   # type: (str, bytes, int
     enc_iter = int.to_bytes(iterations, length=3, byteorder='big', signed=False)
     auth_ver = b'\x01' + enc_iter + salt + derived_key
     return base64_url_encode(auth_ver)
+
+
+def confirm(msg):
```
sk-keeper, line 81: Commander already has a similar method
https://github.com/Keeper-Security/Commander/blob/ff396298ad720006b4ba7a33b9b86549012552af/keepercommander/commands/base.py#L75

sk-keeper (approved):

PR #458, opened 2021-08-30
Fix v2 generate password

idimov-keeper:


keepercommander/commands/recordv3.py
```diff
@@ -655,6 +655,11 @@ def execute(self, params, **kwargs):
                 recordv2.RecordEditCommand().execute(params, **kwargs)
             return
 
+        # if record is v2
+        if rv not in (3, 4):
```
maksimu: how about other versions like ksm app record v5? maybe it is better to say if version <= 2?

maksimu (approved):

PR #474, opened 2021-09-07
Fix ksm (KC-256)

kihneman:
This PR does the following:
- Fix the ksm command by using shlex.split on the command line that is passed to subprocess.check_call (This is only needed on non-Windows platforms)
- Add more info to the error message that is returned if ksm is not installed, i.e. a URL that provides installation instructions for ksm
- Don't display the output of the `which ksm` or `where ksm` that is used to determine if ksm is installed

### DIFF
--- a/keepercommander/constants.py
+++ b/keepercommander/constants.py
@@ -1,6 +1,7 @@
 from datetime import timedelta
 
 
+# Timeout constants
 # Set to default value by using timedelta of 0
 TIMEOUT_DEFAULT = timedelta(0)
 TIMEOUT_MIN = timedelta(minutes=1)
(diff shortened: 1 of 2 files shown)


keepercommander/constants.py
```diff
@@ -111,3 +113,10 @@
 ENFORCEMENTS = {}
 for e in _ENFORCEMENTS:
     ENFORCEMENTS[e[0].lower()] = e[2].lower()
+
+# OS dependent constants
```
maksimu, line 117: how about making functions like this: `isMac` = Mac OS, `isUnix` = Mac OS, Linux, Unix, `isWindows` = Any Windows ?

maksimu (approved):

PR #475, opened 2021-09-08
Add --copy-uid option to clipboard-copy (KC-253)

kihneman:
This PR does the following:
- Update the help for clipboard-copy to show the alias "cc"
- Add the "--copy-uid/-cu" option to copy the record UID
- Add better output to the clipboard-copy to show what was copied to clipboard, i.e. "Password", "Login", or "Record UID"

### DIFF
--- a/keepercommander/commands/recordv3.py
+++ b/keepercommander/commands/recordv3.py
@@ -1479,7 +1479,7 @@ class ClipboardCommand(Command):
                 pyperclip.copy(txt)
                 logging.info(f'{copy_item} copied to clipboard')
             else:
-                print(f'{copy_item}:\n{txt}')
+                print(txt)
             if copy_item == 'Password':
                 params.queue_audit_event('copy_password', record_uid=record_uid)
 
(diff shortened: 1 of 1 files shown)


keepercommander/commands/recordv3.py
```diff
@@ -1463,15 +1464,23 @@ def execute(self, params, **kwargs):
...
         if txt:
             if kwargs['output'] == 'clipboard':
                 import pyperclip
                 pyperclip.copy(txt)
-                logging.info('Copied to clipboard')
+                logging.info(f'{copy_item} copied to clipboard')
             else:
-                print(txt)
-            if not kwargs.get('login'):
+                print(f'{copy_item}:\n{txt}')
```
sk-keeper: I think this change may break customer's script that rely on existing output.
It was a customer's request to dump the password to stdout in order to use it in shell pipeline.

kihneman: I removed the extra text from the copy to stdout.

sk-keeper (approved):

PR #480, opened 2021-09-09
Addressed date parsing traceback in the "created" field

dstromberg:


keepercommander/commands/enterprise.py
```diff
@@ -3027,7 +3027,8 @@ def convert_value(self, field, value, **kwargs):
             return ''
 
         if field == "created":
-            dt = datetime.datetime.utcfromtimestamp(int(value)).replace(tzinfo=datetime.timezone.utc).astimezone(tz=None)
+            parsed_time = datetime.datetime.strptime(value, '%Y-%m-%d %H:%M:%S.%f')
```
sk-keeper: this change breaks
audit-report --report-type=raw 
command

dstromberg: Thanks Sergey.  Are there any other tests I should be running?

sk-keeper: I think that is it.
If we check the type and do not change `int` branch that would be enough.
At least we do not break the existing functionality

sk-keeper (approved):

PR #481, opened 2021-09-10
Added --details option to totp command

dstromberg:


keepercommander/commands/recordv3.py
```diff
@@ -1768,6 +1770,35 @@ def __init__(self, record_uid, record_title, paths):
         self.paths = paths
 
 
+def extract_otp_url(data):
```
sk-keeper: It looks like this code does not work for legacy records

sk-keeper: totp --details  
do not make any difference for me

dstromberg: > It looks like this code does not work for legacy records

True - it currently makes no attempt to deal with legacy records.  Should it?

dstromberg: > totp --details
> do not make any difference for me

You have to specify a record.

dstromberg: I've added support for legacy records.

keepercommander/commands/record_common.py
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env python3
+
+"""Code/Data shared between recordv3.py and record.py."""
+
+import urllib.parse
+
+import pudb
```
sk-keeper: Why this module was introduced into the application?

sk-keeper (approved):

keepercommander/commands/record_common.py
dstromberg: As we slacked about, and as the docstring says, this new file has code shared by record.py and recordv3.py.  It has the advantages of avoiding code duplication, avoiding cyclic imports, and being thoroughly statically analyzed.

dstromberg: Or do you mean urllib.parse?  Or pudb?  urllib.parse is there to extract data from a URL's query string.  pudb was a mistake; I've removed it.

PR #489, opened 2021-09-14
Improve audit report (KC-122)

kihneman:
This PR does the following:
- Adds additional available columns `record_title`, `record_url`, `shared_folder_title`, `team_title`, `node_title`, `role_title` to the audit report. The columns are used by default with the audit report of type `raw`.
- Removes the `--detail` option and functionality in favor of the above additional columns.
- Removes other lookup functions when ids are displayed.


keepercommander/commands/api/compliance_report.py
```diff
@@ -0,0 +1,29 @@
+#  _  __
+# | |/ /___ ___ _ __  ___ _ _ ®
+# | ' </ -_) -_) '_ \/ -_) '_|
+# |_|\_\___\___| .__/\___|_|
+#              |_|
+#
+# Keeper Commander
+# Copyright 2018 Keeper Security Inc.
```
maksimu: update copyright date

keepercommander/commands/helpers/audit_report.py
```diff
@@ -0,0 +1,69 @@
...
+# | |/ /___ ___ _ __  ___ _ _ ®
+# | ' </ -_) -_) '_ \/ -_) '_|
+# |_|\_\___\___| .__/\___|_|
+#              |_|
+#
+# Keeper Commander
+# Copyright 2021 Keeper Security Inc.
+# Contact: ops@keepersecurity.com
+#
+from dataclasses import dataclass
```
sk-keeper: Do we really need to introduce a new dependency?

kihneman: The reason is that I am intentionally trying to break up these very long modules into smaller pieces. I personally feel that most of our files are way too long and need to be broken up. I would be happy to discuss this matter, though. Also if you want me to put all of these back into the main file I will do it.

sk-keeper: That's fine. Why we need to introduce a new dependency?
Why it cannot be done without adding `dataclasses` dependency?

kihneman: `dataclasses` are just a nice new feature of Python which are standard as of Python 3.7. It makes data objects more readable because the items of the data object have names unlike storing in a tuple or list. Another alternative would be to use a `namedtuple`. `dataclasses` is only needed in  the `install_requires` of the setup.py file for Python < 3.7.

kihneman: I suppose that I am not using that many of the dataclass features and can remove the dependency if you like. Mainly it eliminates the need for the `__init__`.

sk-keeper: These dependency change seems to split Commander running environment into 2 parts
pre 3.7 and 3.7 and up

Who is going to test it for Python 3.6?

kihneman: The automated CI testing tests run in Python 3.6 and Python 3.9. However, our tests are not that thorough. It wouldn't be that hard to change the code to work without dataclasses. We should take note, though, that Python 3.6 is only receiving support and security updates through the end of this year. After that, anyone using Python would be advised to upgrade to an officially supported version of Python.

kihneman: I removed the dataclasses dependency

sk-keeper (approved):

Issue #488, opened 2021-09-14
--node option not limiting results in ei with -u, -t or -r arguments


carlosduplar:
Following #482 , Commander stopped breaking with `ei --node="nodename"` commands. 

Command is not behaving properly, though: --node="nodename" is not filtering results when used with `-u`, `-t` or `-r` arguments. `-n` is currently the only argument correctly filtered by --node.

sk-keeper:
This issue will be fixed in the next Commander release.

PR #492, opened 2021-09-15
enterprise-info command to respect --node parameter

sk-keeper:


maksimu (approved):
Tested, works great
closes issue #488

PR #497, opened 2021-09-16
Import/Export Records with Types. KC-229

sk-keeper:


keepercommander/api.py
```diff
@@ -5,7 +5,7 @@
 #              |_|            
 #
 # Keeper Commander 
-# Copyright 2017 Keeper Security Inc.
+# Copyright 2021 Keeper Security Inc.
```
dstromberg, line 8: FWIW, ISTR that a company should update their copyright notices for appropriate years, annually.

dstromberg (commented):
How are you testing this?

I wish there were more docstrings and comments.

The identifiers look pretty good.

keepercommander/importer/importer.py
```diff
@@ -138,17 +141,56 @@ def get_folder_path(self):
...
+            return f'${self.type.lower()}.{self.label.lower()}'
+        elif self.type:
+            return f'${self.type.lower()}'
+        else:
+            return (self.label or '').lower()
+
+    def hash_key(self):
+        value = self.value
+        if value:
+            if type(value) is str:
```
dstromberg, line 161: No big deal, but this could be written as if isinstance(value, str) or if isinstance(value, (str, ))
pylint prefers isinstance, but I've turned that off in our pylintrc.
It might be good to be consistent.

dstromberg:
I wish we had smaller code units.  EG, radon cc gives convert_keeper_record and _import 'F' grades for being too complex.

keepercommander/importer/imp_exp.py
```diff
@@ -384,11 +724,57 @@ def execute_import_folder_record(params, folders, records):
             rs_record.extend(import_rs.recordResponse)
 
     report_statuses('folder', (element.status for element in rs_folder))
-    report_statuses('record', (element.status for element in rs_record))
+    report_statuses('legacy record', (element.status for element in rs_record))
 
     return rs_folder, rs_record
 
 
+def record_status_to_str(status):  # type: (record_proto.RecordModifyStatus) -> str
```
dstromberg, line 732: Again, not a big deal, but this could be done a bit faster with collections.default_dict.  Readability is probably a tossup.

keepercommander/importer/imp_exp.py
```diff
@@ -678,211 +1064,269 @@ def prepare_folder_add(params, folders, records):
...
+
+def value_to_token(value): # type: (any) -> str
+    if not value:
+        return ''
+    if type(value) == str:
+        return value
+    if type(value) == list:
+        return ','.join((value_to_token(x) for x in value))
+    if type(value) == dict:
+        pairs = [x for x in value.items()]
```
dstromberg, line 1098: FWIW, this could probably be simplified a little to pairs = list(value.items())

keepercommander/importer/imp_exp.py
```diff
@@ -678,211 +1064,269 @@ def prepare_folder_add(params, folders, records):
...
-    """
-    Turn a preexisting record into an iterable of str's for hashing.
-
-    Examine just the relevant parts of the record.
-
-    For now at least, this is the same as tokenize_partial_import_record().
-    """
-    return tokenize_partial_import_record(record)
+    record_type = record.type or ""
+    yield f'$type:{record_type}'
```
dstromberg, line 1111: What is the meaning of all these dollar signs?  This is more of a cross-training thing. It probably merits a comment though.

keepercommander/importer/importer.py
```diff
@@ -138,17 +141,56 @@ def get_folder_path(self):
...
+                pass
+            elif type(value) is list:
+                value = '|'.join((str(x) for x in value))
```
dstromberg, line 164: A tiny thing, but I don't think it's necessary to double the parens here.  The following should work fine:
value = '|'.join(str(x) for x in value)

keepercommander/importer/keepass/keepass.py
```diff
@@ -186,7 +175,21 @@ def do_import(self, filename):
                                 elif key == 'Notes':
                                     record.notes = value
                                 else:
-                                    record.custom_fields[key] = value
+                                    if key[0] == '$':
+                                        pos = key.find(':')
```
dstromberg, line 179: A lot of uses of str.find can be simplified with, EG:
>>> s = 'abc:def:ghi'
>>> s.partition(':')
('abc', ':', 'def:ghi')

There's also an rpartition method for starting from the other end of the string.

dstromberg (commented):
I probably should test before approving.  But how?

keepercommander/api.py
sk-keeper: Commander is the first and only public source code repository. There are no established rules on updating copyright.
Developers usually modify the header when they touch the file

keepercommander/importer/importer.py
sk-keeper: I agree. I checked the difference between isinstance and type. It is stated that is isinstance is preferable.
I started using isinstance.

keepercommander/importer/imp_exp.py
sk-keeper: Python can compete with Perl on `There's more than one way to do it / TMTOWTDI`

keepercommander/importer/imp_exp.py
sk-keeper: `My Vault> rti -lr login                                                                                                                                                                                                            
  Record Type ID  Content
----------------  ----------------------------------
               1  {
                    "$id": "login",
                    "categories": [
                      "login"
                    ],
                    "description": "Login template",
                    "fields": [
                      {
                        "$ref": "login"
                      },
                      {
                        "$ref": "password"
                      },
                      {
                        "$ref": "url"
                      },
                      {
                        "$ref": "fileRef"
                      },
                      {
                        "$ref": "oneTimeCode"
                      }
                    ]
                  }
`

sk-keeper: For some reason, it was decided to mark types for "record V3" with $

keepercommander/importer/importer.py
sk-keeper: Thanks. It is good to know.

keepercommander/importer/keepass/keepass.py
sk-keeper: Yes, it makes code better. I did not know about `partition`

`field_type, _, field_label = s.partition(':')`

keepercommander/api.py
sk-keeper: I personally find docstrings and comments distracting.
Type hints and lack of IDE warnings are more helpful for me.

keepercommander/importer/imp_exp.py
sk-keeper: I did not know about defaultdict in python. Thanks

PR #500, opened 2021-09-17
Automator name is used instead of Node name

sk-keeper:


maksimu (approved):

dstromberg:
It looks reasonable.  Looks like it's been merged too.  :)

PR #510, opened 2021-09-27
Fix lastpass import of records without title

kihneman:
If a Lastpass record being imported has no title then the import will crash. ~This PR adds in a default of "Lastpass Import Untitled [count]" for each untitled record being imported.~

The records that didn't have titles had "http://group" in the URL because they were folders. This PR skips those items.


kihneman:
Don't merge this PR. Updates are needed.

sk-keeper (approved):

sk-keeper (approved):

PR #514, opened 2021-09-28
Import shared folder members with shared folder (KC-286)

kihneman:
This PR does the following:
- Adds class LastpassSharedFolder to store shared folder name with permissions for shared folder imported from Lastpass.
- Makes API call to Lastpass to retrieve shared folder permissions which are stored in Vault.shared_folders as list of LastpassSharedFolder.
- Yield list of importer.SharedFolder to the Keeper import with permissions populated from the lastpass shared folders.

### DIFF
--- a/keepercommander/importer/lastpass/fetcher.py
+++ b/keepercommander/importer/lastpass/fetcher.py
@@ -58,8 +58,12 @@ def fetch_shared_folder_members(session, shareid, web_client=http):
     if response.status_code != requests.codes.ok:
         raise NetworkError()
 
-    # We're only returning the users that have access to the share and not the groups
-    shared_folder_members = json.loads(response.content.decode('utf-8'))['users']
+    response_dict = json.loads(response.content.decode('utf-8'))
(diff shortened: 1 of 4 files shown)


keepercommander/importer/lastpass/fetcher.py
```diff
@@ -40,15 +42,31 @@ def logout(session, web_client=http):
...
         raise NetworkError()
 
     return blob.Blob(decode_blob(response.content), session.key_iteration_count)
 
 
+def fetch_shared_folder_members(session, shareid, web_client=http):
+    url = 'https://lastpass.com/getSharedFolderMembers.php' + query_string + f'&shareid={shareid}'
+    response = web_client.get(url, cookies={'PHPSESSID': session.id})
+
+    if response.status_code != requests.codes.ok:
```
sk-keeper, line 58: I think we need to be sure that this code does not throw any error.
It's better this method does nothing that cancels the whole import

sk-keeper (approved):

PR #519, opened 2021-09-29
KSM: init config and token with host

maksimu:
- new flag to generate new client that will generate config in JSON or Base64

- one time tokens now contain hostname


sk-keeper (approved):

idimov-keeper (approved):
maybe we should add a warning when using --config-init without --unlock-ip

idimov-keeper (approved):

PR #559, opened 2021-10-12
Import: strip URL for record audit

sk-keeper:


keepercommander/utils.py
```diff
@@ -84,6 +85,14 @@ def create_auth_verifier(password, salt, iterations):   # type: (str, bytes, int
     return base64_url_encode(auth_ver)
 
 
+def url_strip(url):   # type: (str) -> str
+    try:
+        result = urlparse(url)
+        return result.netloc + result.path
```
idimov-keeper, line 91: that doesn't match the [technical specification](https://keeper.atlassian.net/wiki/spaces/KA/pages/991428847/records+update) according to which we should also preserve any fragment(s) - ex. **records_update**: "The URL is stripped of its **protocol** and **query** parameters" ONLY!

https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#/media/File:URI_syntax_diagram.svg




https://en.wikipedia.org/wiki/Uniform_Resource_Identifier

PR #589, opened 2021-10-27
Skip http://group records before checking for missing title

kihneman:
This is another fix for checking for LastPass records without a title. There are records from LastPass with a URL `http://group` and no title that we skip anyway. These records should be skipped before checking for a missing title.


kihneman:
This issue is now being fixed in #593

PR #708, opened 2022-03-30
Add kcm command for making connections through Glyptodon

kihneman:
This PR adds a command `kcm`. The command currently requires the following to make a Glyptodon connection:
- The option `--json-auth` (`-j`) which specifies a record with the shared encryption key stored in the password field and the URL of the Glyptodon server stored in the Website Address (Login URL) field.
- One or more records of type `Server` that also include a custom field named `Protocol`. These records should specify the hostname and port in the Hostname field, the login name, the password, and either `ssh` or `rdp` in the custom Protocol field.


kihneman:
Closing this PR since we are not ready for this feature to be available yet.

PR #726, opened 2022-04-20
Display folder. Include parent and shared folder information

sk-keeper:


maksimu (approved):

keepercommander/commands/recordv3.py
```diff
@@ -2230,70 +2234,6 @@ def execute(self, params, **kwargs):
             raise CommandError('get', 'Record is not version 3 (record type)')
 
         fmt = kwargs.get('format') or 'detail'
-
```
idimov-keeper, line 2233: Is this intentional - adding verbosity removes the most verbose section(s)

sk-keeper: I tried to remove the dead code. It seems that recordv3 was a full copy of recordv2. recordv3 get command displays only v3 records

PR #769, opened 2022-06-15
Feature/action report cmd (reporting only)

aaunario-keeper:


keepercommander/commands/aram.py
```diff
@@ -1510,3 +1518,61 @@ def get_password_data():
...
+                'report_type': 'span',
+                'scope': 'enterprise',
+                'columns': [username_field, 'audit_event_type'],
+                'filter': {
+                    'audit_event_type': ['login'] if event_types is None else event_types,
+                    'created': {'min': int(dt.timestamp())}
+                },
+                'aggregate': ['last_created'],
+                'limit': 1000,
+                'group_by': ['last_created', 'audit_event_type', username_field]
```
sk-keeper: documentation was incorrect. "group_by" is not a valid parameter

keepercommander/commands/aram.py
```diff
@@ -1510,3 +1518,61 @@ def get_password_data():
...
+
+    def execute(self, params, **kwargs):
+        def get_request(days_since=30, event_types=None, username_field='username'):
+            days_since = 30 if not isinstance(days_since, int) else days_since
+            dt = datetime.datetime.now() - datetime.timedelta(days=days_since)
+            rq = {
+                'command': 'get_audit_event_reports',
+                'report_type': 'span',
+                'scope': 'enterprise',
+                'columns': [username_field, 'audit_event_type'],
```
sk-keeper: event type is probably not needed here.

keepercommander/commands/aram.py
aaunario-keeper: fixed in latest commit

keepercommander/commands/aram.py
aaunario-keeper: fixed in latest commit to this branch

PR #781, opened 2022-06-28
Update enterprise.py

securitytime:
For showing roles, supported_columns was incorrectly set to SUPPORTED_TEAM_COLUMNS instead of SUPPORTED_ROLE_COLUMNS.
This caused enterprise-info --roles to omit columns in the output.


aaunario-keeper (commented):
change looks good, though you should also change the logging text in line 750 from "Supported **team** columns:" to "Supported **role** columns:"

aaunario-keeper:
closing this PR and adding relevant changes to another PR

PR #783, opened 2022-06-28
Update enterprise.py

sk-keeper:
For showing roles, supported_columns was incorrectly set to SUPPORTED_TEAM_COLUMNS instead of SUPPORTED_ROLE_COLUMNS.
This caused enterprise-info --roles to omit columns in the output.


aaunario-keeper (commented):
looks good

PR #801, opened 2022-07-19
new python requirement now >= ver. 3.7

aaunario-keeper:


kihneman (changes requested):
Please add this change to the other PR with updates from 3.6 to 3.7

kihneman (approved):
Sorry I was confused. This PR is good

PR #844, opened 2022-09-16
Release

sk-keeper:


keepercommander/commands/utils.py
```diff
@@ -1495,8 +1495,19 @@ def add_client(params, app_name_or_uid, count, unlock_ip, first_access_expire_on
...
+        elif isinstance(unlock_ip, str) and len(unlock_ip) > 0:
+            unlock_ip = unlock_ip.lower()
+            if unlock_ip == 't':
+                unlock_ip = 'true'
+            elif unlock_ip == 'y':
+                unlock_ip = 'yes'
+            elif unlock_ip == '1':
+                unlock_ip = 'true'
+            is_ip_unlocked = unlock_ip in ['yes', 'true']
         else:
```
maksimu, line 1509: we have similar login [HERE](https://github.com/Keeper-Security/Commander/blob/60a13e7944746ab7deada08af576ef384e642dae/keepercommander/commands/utils.py#L468-L473). Maybe it can be combined in a special function?

PR #946, opened 2023-02-10
user-report fixes (for KC-519):

aaunario-keeper:
1) no-limit look-back-days 
2) infinite loop bug w/ empty usernames filter 
3) most-recent logins 
4) no-rows-returned bug

### DIFF
--- a/keepercommander/commands/enterprise.py
+++ b/keepercommander/commands/enterprise.py
@@ -3176,20 +3176,16 @@ class UserReportCommand(EnterpriseCommand):
                     self.user_teams[tu['enterprise_user_id']].append(self.teams[tu['team_uid']])
 
         limit = API_EVENT_SUMMARY_ROW_LIMIT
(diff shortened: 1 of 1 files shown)


keepercommander/commands/enterprise.py
```diff
@@ -3176,20 +3176,16 @@ def execute(self, params, **kwargs):
                     self.user_teams[tu['enterprise_user_id']].append(self.teams[tu['team_uid']])
 
         limit = API_EVENT_SUMMARY_ROW_LIMIT
-        look_back_days = kwargs.get('days') or 365
-        to_ts = int(datetime.datetime.now().timestamp())
+        look_back_days = kwargs.get('days', 365)
```
idimov-keeper, line 3179: what if the key is present but with value `None`? Looks like we still need `or 365` part

sk-keeper (approved):

PR #976, opened 2023-03-16
 Should Do Filtering By Record UID. KC-544

sk-keeper:


keepercommander/commands/aram.py
```diff
@@ -1584,78 +1584,75 @@ def clean_up():
...
+            events = rs['audit_event_overview_report_rows']
+            done = len(events) < API_EVENT_SUMMARY_ROW_LIMIT - 100
+            if not done:
+                from_date = events[-1]['last_created']
+                if events[0]['last_created'] == from_date:
+                    from_date += 1
+            for event in events:
+                record_uid = event['record_uid']
+                ra = None    # type: Optional[StorageRecordAging]
+                if record_uid not in changed_records:
```
aaunario-keeper, line 1627: given the "ascending" order specified in the "get_audit_event_reports" request object, this means only the OLDEST record-add/record-password-change event is included in the report

keepercommander/commands/aram.py
```diff
@@ -1584,78 +1584,75 @@ def clean_up():
...
+            'created': {'min': 0}
+        }
+        rq = {
+            'command':      'get_audit_event_reports',
+            'scope':        'enterprise',
+            'report_type':  'span',
+            'columns':      ['record_uid', 'audit_event_type'],
+            'aggregate':    ['last_created'],
+            'filter':       audit_filter,
+            'order':        'ascending',
```
aaunario-keeper, line 1607: "descending" order should be used instead to ensure that the most recent record-add/ password-change event dates are used

keepercommander/commands/aram.py
```diff
@@ -1584,78 +1584,75 @@ def clean_up():
...
+        logging.info('Loading record password change information...')
+        while not done:
+            audit_filter['created']['min'] = from_date
+            rs = api.communicate(params, rq)
+            events = rs['audit_event_overview_report_rows']
+            done = len(events) < API_EVENT_SUMMARY_ROW_LIMIT - 100
+            if not done:
+                from_date = events[-1]['last_created']
+                if events[0]['last_created'] == from_date:
+                    from_date += 1
```
aaunario-keeper, line 1623: should be `from_date -= 1` (given the "descending" order change suggested)

keepercommander/commands/aram.py
```diff
@@ -1584,78 +1584,75 @@ def clean_up():
...
+            'order':        'ascending',
+            'limit':        API_EVENT_SUMMARY_ROW_LIMIT
+        }
+        changed_records = {}    # type: Dict[str, StorageRecordAging]
+        last_added = 0
+        last_password_change = 0
+        done = False
+        logging.info('Loading record password change information...')
+        while not done:
+            audit_filter['created']['min'] = from_date
```
aaunario-keeper, line 1616: the upper limit (and not the lower limit) should be adjusted upon each subsequent request (instead of setting the "min" of the "created" property, we should be setting the "max" to the least recent timestamp of the last set of returned event rows)

keepercommander/commands/aram.py
```diff
@@ -1584,78 +1584,75 @@ def clean_up():
...
+            rs = api.communicate(params, rq)
+            events = rs['audit_event_overview_report_rows']
+            done = len(events) < API_EVENT_SUMMARY_ROW_LIMIT - 100
```
aaunario-keeper, line 1619: we should probably add an additional check (by checking "to_date" value) to avoid unnecessary API requests

aaunario-keeper (approved):

PR #1181, opened 2024-02-05
(KC-473) `security-audit-report --breachwatch` improvement:

aaunario-keeper:
added `bw report` as alias for command

### DIFF
--- a/keepercommander/commands/breachwatch.py
+++ b/keepercommander/commands/breachwatch.py
@@ -15,6 +15,7 @@ import getpass
 import logging
 from typing import Optional, Any, Dict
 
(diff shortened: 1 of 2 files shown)


keepercommander/commands/breachwatch.py
```diff
@@ -15,6 +15,7 @@
 import logging
 from typing import Optional, Any, Dict
 
+from .security_audit import SecurityAuditReportCommand
```
sk-keeper, line 18: Would it be possible to have a command alias like this?
https://github.com/Keeper-Security/Commander/blob/17c88e1d4aded9d1aa345ef592e939415a96d4a9/keepercommander/commands/security_audit.py#L23

PR #1184, opened 2024-02-09
(KC-740) `share-folder` fix:

aaunario-keeper:
1) added API request-batching to allow updating permissions for shared folders w/ > 500 users or records (avoiding KA exception on requests that exceed that limit)


aaunario-keeper:
Hold off on reviewing this while I fix this bug

PR #1222, opened 2024-04-03
improvement: option to raise errors for query_enterprise

Silentapproach:
Currently, any error raised inside query_enterprise is only logged as a warning, with no easy way to catch it in an except clause outside the function.

Because I didn't want my code to break any codebases, it is implemented as an optional parameter.

### DIFF
--- a/keepercommander/api.py
+++ b/keepercommander/api.py
@@ -1273,7 +1273,7 @@ def get_record_shares(params, record_uids, is_share_admin=False):
         return result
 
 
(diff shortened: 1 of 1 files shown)


sk-keeper:
`api.query_enterprise` can be replaced with `enterprise.query_enterprise` to get the same results.

Silentapproach:
@sk-keeper yep that worked. Thanks for the tip. I leave it up to you, if that gets merged or not. I mean it doesn't hurt anybody :-)

PR #1234, opened 2024-04-30
Typo "Sharable" instead of "Shareable"

Silentapproach:
Error Shareable
When running the command ShareRecordCommand.execute on a record that is already shared to the user that it should be shared with again, an error occurs.

The error states that there is no such key as "sharable" because the correct key name is "shareable".

Error: 

keepercommander\commands\register.py", line 884, in execute
    ro.shareable = True if can_share else current['sharable']
                                          ~~~~~~~^^^^^^^^^^^^
KeyError: 'sharable'

When I searched through the whole project I found more than two results (the ones addressed in this pr). 
I didn't want to touch them because I only worked with ShareRecordCommand out of the ones that contain "sharable"

### DIFF
--- a/keepercommander/commands/register.py
+++ b/keepercommander/commands/register.py
@@ -881,13 +881,13 @@ class ShareRecordCommand(Command):
                             transfer_ruids.add(record_uid)
                         else:
                             ro.editable = True if can_edit else current['editable']
-                            ro.shareable = True if can_share else current['sharable']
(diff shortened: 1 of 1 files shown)


aaunario-keeper:
Closing this PR, as this issue has been addressed in release v16.10.14 (available [here](https://github.com/Keeper-Security/Commander/releases/tag/v16.10.14))

PR #1335, opened 2024-12-13
CyberArk Account import functionality

amigus:
Add a module that imports accounts from CyberArk using the REST API

### DIFF
--- a/keepercommander/importer/cyberark/cyberark.py
+++ b/keepercommander/importer/cyberark/cyberark.py
@@ -1,0 +1,103 @@
+import getpass
+import urllib3
+import requests
+
+urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
+
+
+from ..importer import BaseImporter, Record, RecordField
+
+
+class CyberArkImporterException(BaseException):
+    def __init__(self, response, message):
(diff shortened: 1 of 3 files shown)


keepercommander/importer/cyberark/cyberark.py
```diff
@@ -0,0 +1,103 @@
...
+    # Request timeout in seconds
+    TIMEOUT = 10
+
+    @classmethod
+    def get_url(cls, pvwa_host, endpoint):
+        return f"https://{pvwa_host}/PasswordVault/API/{cls.ENDPOINTS[endpoint]}"
+
+    def do_import(self, filename, **kwargs):
+        # The CyberArk API implements paging with a limit of 1000 records per page
+        params = {"limit": 1000, "offset": 0}
```
sk-keeper: Commander uses `params` variable name for its main KeeperParams instance in the entire code base.
It would be nice to keep this name reserved for that purpose.

sk-keeper (approved):

keepercommander/importer/cyberark/cyberark.py
amigus: Done along with some improvements.

PR #1365, opened 2025-01-22
Changed pam action rotate output to print gateway in use

idimov-keeper:


keepercommander/commands/discoveryrotation.py
```diff
@@ -2134,7 +2134,7 @@ def execute(self, params, **kwargs):
             encrypted_transmission_key=encrypted_transmission_key,
             encrypted_session_token=encrypted_session_token)
 
-        print_router_response(router_response, 'job_info', conversation_id)
+        print_router_response(router_response, 'job_info', conversation_id, gateway_uid=gateway_uid)
```
sk-keeper, line 2137: I can see this function is called from 9 commander's commands
only one command adds gateway parameter `pam action rotate`

Some other commands also have gateway context like `pam action gateway-info`

Are we sure this fix is not needed in other commands that shows scheduled commands?

idimov-keeper: Only rotate command auto-detects gateway from record (hence the need to print it in output) - remaining commands either already know it or don't need it, so no need to print it out.
_BTW: I also fixed all broken invocations of `print_router_response` that didn't take into account function change (and newly added `response_type` parameter)_

sk-keeper (approved):

PR #1371, opened 2025-02-04
Feat/search by name

nikitaart2805:
Feature Description
We introduced a new --field parameter that lets users restrict searches to specific record fields. For example, specifying --field title searches only in the record’s title, while --field notes focuses on notes. This flexibility makes searches more precise when you know exactly which field(s) to filter on.

Benefit
By allowing users to target fields rather than searching across all of them, the new feature avoids returning matches from irrelevant fields. This results in fewer, more relevant records in the output and creates a more efficient user experience.

How It Works

General Pattern Search:
By default, the CLI searches across multiple fields for any provided pattern. This can sometimes return broad or cluttered results if the pattern appears in notes, descriptions, or other fields.
Field-Based Search:
When the user specifies one or more --field parameters (e.g., --field title, --field notes), the search logic applies the pattern only to those named fields. This filters out records where the pattern only appears in other fields, delivering a more focused set of results.
Overall Impact

Reduced Noise: Users see fewer extraneous matches, as only the specified fields are searched.
Improved Flexibility: Users can pass multiple --field flags for more nuanced searches (e.g., --field title --field notes).
Quicker Targeting: When searching by title only, it’s easier to locate a known record by name without sifting through other field matches.
Example

**General Search: Running list AWS might show multiple records if they have “AWS” in any field.
Field-Based Search: Running  list --field title AWS finds only records whose titles contain “AWS,” ignoring notes and other fields.**

Screenshot 2025-02-03 at 8 22 13 PM

### DIFF
--- a/keepercommander/commands/record.py
+++ b/keepercommander/commands/record.py
@@ -106,7 +106,7 @@ list_parser.add_argument('--output', dest='output', action='store',
                          help='output file name. (ignored for table format)')
 list_parser.add_argument('-t', '--type', dest='record_type', action='append',
                          help='List records of certain types. Can be repeated')
-list_parser.add_argument('--name', dest='name', type=str, help='Filter records by name')
+list_parser.add_argument('--field', dest='field', action='append', help='Filter records by specific field(s). Can be specified multiple times.')
 list_parser.add_argument('pattern', nargs='?', type=str, action='store', help='search pattern')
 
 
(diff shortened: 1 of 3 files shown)


nikitaart2805:
@sk-keeper Take another look, please :) thank u !

sk-keeper:
@nikitaart2805  Thank you for your contribution to the Commander.
Your changes will be included into the next release.

PR #1548, opened 2025-08-07
Implement audit-alert enable/disable with --all flag

pvagare-ks:


keepercommander/commands/audit_alerts.py
```diff
@@ -843,6 +855,138 @@ def execute(self, params, **kwargs):
...
+                rq = {
+                    'command': 'put_enterprise_setting',
+                    'type': 'AuditAlertContext',
+                    'settings': {
+                        'id': alert_id,
+                        'disabled': False
+                    }
+                }
+                api.communicate(params, rq)
+                enabled_count += 1
```
satish-metron:
```suggestion
        requests = []
        for alert in alert_filter:
            alert_id = alert.get('id')
            if alert_id:
                rq = {
                    'command': 'put_enterprise_setting',
                    'type': 'AuditAlertContext',
                    'settings': {
                        'id': alert_id,
                        'disabled': False
                    }
                }
                requests.append(rq)
        
        if not requests:
            logging.info('No valid alerts found to enable')
            return
        
        logging.info(f'Enabling {len(requests)} alert(s) in parallel...')
        results = api.execute_batch(params, requests)
        
        enabled_count = sum(1 for result in results if result.get('result') == 'success')
```

Hey maybe you can do it this way, you will be able to parallelize some requests as `exectute_batch` option is already prebuilt into commander

keepercommander/commands/audit_alerts.py
```diff
@@ -843,6 +855,138 @@ def execute(self, params, **kwargs):
...
+        alerts = self.load_settings(params)
+        if not isinstance(alerts, dict):
+            logging.info('No alerts found')
+            return
+        alert_filter = alerts.get('AuditAlertFilter')
+        if not isinstance(alert_filter, list):
+            logging.info('No alerts found')
+            return
+        
+        disabled_count = 0
```
satish-metron: Batch this similarly as shown in enable_all_alerts functionality

keepercommander/commands/audit_alerts.py
```diff
@@ -843,6 +855,138 @@ def execute(self, params, **kwargs):
...
+            if not target:
+                raise CommandError('Alert target is required unless using --all flag')
+            self._disable_single_alert(params, target, kwargs)
+    
+    def _disable_single_alert(self, params, target, kwargs):
+        alert = AuditSettingMixin.get_alert_configuration(params, target)
+
+        rq = {
+            'command': 'put_enterprise_setting',
+            'type': 'AuditAlertContext',
```
satish-metron: I see that we are using this string in many places, so maybe make it as constant and use it everywhere ? that way there will not be scope for mistakes

keepercommander/commands/audit_alerts.py
```diff
@@ -843,6 +855,138 @@ def execute(self, params, **kwargs):
...
+        target = kwargs.get('target')
+        
+        if enable_all:
+            if target:
+                raise CommandError('Cannot specify both alert target and --all flag')
+            self._enable_all_alerts(params)
+        else:
+            if not target:
+                raise CommandError('Alert target is required unless using --all flag')
+            self._enable_single_alert(params, target, kwargs)
```
satish-metron: This conditional can be simplified here and in disabled_all case too.

Something like this can be used to make code understandable. Please check if my suggestion covers all your cases and feel free to change if you feel like it.
```suggestion
        if enable_all and target:
            raise CommandError('Cannot specify both alert target and --all flag')
        elif enable_all:
            self._enable_all_alerts(params)
        elif not target:
            raise CommandError('Alert target is required unless using --all flag')
        else:
            self._enable_single_alert(params, target, kwargs)
```

keepercommander/commands/audit_alerts.py
pvagare-ks: Done

keepercommander/commands/audit_alerts.py
pvagare-ks: Done

keepercommander/commands/audit_alerts.py
pvagare-ks: Done

keepercommander/commands/audit_alerts.py
pvagare-ks: Okay

PR #1550, opened 2025-08-08
Changes biometric login info from print to logging

dcfSec:
Hi

This pull request changes the biometric login handling from just printing the information to the log handler.

This change has following benefits:

-  you can now pipe the output e.g. from a keeper get into another tool without getting also the biometric login information (stout vs now sterr)
- informational output is now streamlined with other warnings and information


craiglurey (approved):
Approved thank you. Merging into release branch.

PR #1603, opened 2025-09-24
Normalize server URL to handle servers with or without scheme/port

sdubey-ks:
**Title:** Normalize Server URLs by handling full server URLs and bare hostnames

**Description:**
This PR improves the generation of vault-related server URLs by standardizing how `params.server` is handled. Previously, if `params.server` included a scheme (e.g., `https://example.com`) or a port, the generated URLs could be malformed.

**Changes include:**

1. **Normalize `params.server` for `keeper_url`**

   * Uses `urlparse` to extract `netloc` if a scheme is present; falls back to `path` for plain hostnames.
   * Constructs `keeper_url` consistently as `https://<server>/vault/email/<email>`.

2. **Normalize `params.server` for vault share URL**

   * Applies the same parsing logic to `params.server` when constructing vault share URLs.
   * Ensures URLs like `https://<server>/vault/share/<encoded_client_key>` are correctly formed regardless of input format.

3. **Unified parsing logic**

   * Both URLs now use the same approach: `parsed = urlparse(params.server)` → `server_netloc = parsed.netloc or parsed.path`.
   * Prevents issues with bare hostnames, URLs with schemes, ports, or trailing slashes.

**Benefits:**

* Prevents malformed URLs due to inconsistent server input.
* Maintains consistent HTTPS URLs for all vault operations.
* Handles edge cases such as `example.com`, `https://example.com`, `example.com:8443`, and URLs with extra paths.

### DIFF
--- a/keepercommander/commands/register.py
+++ b/keepercommander/commands/register.py
@@ -2394,7 +2394,11 @@ class OneTimeShareCreateCommand(Command):
                 query = 'editable=true'
 
             api.communicate_rest(params, rq, 'vault/external_share_add', rs_type=APIRequest_pb2.Device)
-            url = urlunparse(('https', params.server, '/vault/share/', None, query, utils.base64_url_encode(client_key)))
+            # Extract hostname from params.server in case it contains full URL with protocol
+            from urllib.parse import urlparse
+            parsed = urlparse(params.server)
+            server_netloc = parsed.netloc if parsed.netloc else parsed.path  # parsed.path for plain hostname
+            url = urlunparse(('https', server_netloc, '/vault/share/', None, query, utils.base64_url_encode(client_key)))
             urls[record_uid] = str(url)
 
         if params.batch_mode:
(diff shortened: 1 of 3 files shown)


keepercommander/commands/register.py
```diff
@@ -2394,7 +2394,11 @@ def execute(self, params, **kwargs):
                 query = 'editable=true'
 
             api.communicate_rest(params, rq, 'vault/external_share_add', rs_type=APIRequest_pb2.Device)
-            url = urlunparse(('https', params.server, '/vault/share/', None, query, utils.base64_url_encode(client_key)))
+            # Extract hostname from params.server in case it contains full URL with protocol
+            from urllib.parse import urlparse
+            parsed = urlparse(params.server)
+            server_netloc = parsed.netloc if parsed.netloc else parsed.path  # parsed.path for plain hostname
+            url = urlunparse(('https', server_netloc, '/vault/share/', None, query, utils.base64_url_encode(client_key)))
```
sgaddala-ks, line 2401:
```suggestion
            url = urlunparse(('https', server_netloc, '/vault/share', None, query, utils.base64_url_encode(client_key)))
```

sdubey-ks: Removing the trailing slash will break the share URL construction. For example:
- Expected: https://qa.keepersecurity.com/vault/share/#link
- Actual output: https://qa.keepersecurity.com/vault/share#link

PR #1612, opened 2025-09-29
Add --file flag for logging debug logs

pvagare-ks:
## Summary

This PR introduces support for a new `--file` flag to the `debug` command and `this-device` will show where is config.json file in Keeper Commander.  
The flag allows users to send all DEBUG-level logs to a specified file.

---

## Changes

- **New flag:** `debug --file=/path/to/log`
  - Creates (or switches to) a file handler that captures **all DEBUG logs**.
  - Console output level is preserved (INFO/WARNING), so enabling file logging does not spam the terminal with debug lines.
  
 - **config.json file**: `this-device`
   - Now `this-device` command will show config.json file location.

---

## Usage Examples

```bash
debug --file=/tmp/debug.log
# => enables debug logging to /tmp/debug.log

### DIFF
--- a/keepercommander/service/commands/security_config_handler.py
+++ b/keepercommander/service/commands/security_config_handler.py
@@ -32,7 +32,7 @@ class SecurityConfigHandler(Command):
         config_data["is_advanced_security_enabled"] = (
             self.service_config._get_yes_no_input(self.messages['advanced_security_prompt'])
         )
-        config_data["ip_allowed_list"] = '0.0.0.0/0'
+        config_data["ip_allowed_list"] = '0.0.0.0/0,::/0'
         if config_data["is_advanced_security_enabled"] == "y":
             self._configure_advanced_security(config_data)
 
--- a/keepercommander/service/commands/service_config_handlers.py
+++ b/keepercommander/service/commands/service_config_handlers.py
@@ -47,7 +47,7 @@ class ServiceConfigHandler:
     @debug_decorator
     def handle_streamlined_config(self, config_data: Dict[str, Any], args, params: KeeperParams) -> None:
         if args.allowedip is None:
-            args.allowedip = '0.0.0.0/0'
+            args.allowedip = '0.0.0.0/0,::/0'
         
         run_mode = args.run_mode if args.run_mode is not None else "foreground"
         if args.run_mode is not None and run_mode not in ['foreground', 'background']:

### CODE REVIEW

File: keepercommander/cli.py [hunk 1]
```diff
@@ -433,6 +437,98 @@ def force_quit():
...
+        try:
+            os.makedirs(os.path.dirname(os.path.abspath(file_path)), exist_ok=True)
+            
+            for h in list(self.logger.handlers):
+                if getattr(h, '_debug_file', False):
+                    self.logger.removeHandler(h)
+                    try:
+                        h.close()
+                    except:
+                        pass
```
sali-ks: Empty catch block silently ignores errors. Use proper exception

keepercommander/cli.py
```diff
@@ -433,6 +437,98 @@ def force_quit():
...
+        level = logging.DEBUG if enabled else (logging.WARNING if batch_mode else logging.INFO)
+        for h in self.logger.handlers:
+            if isinstance(h, logging.StreamHandler) and not getattr(h, '_debug_file', False):
+                h.setLevel(level)
+    
+    def setup_file_logging(self, file_path):
+        """Setup debug file logging."""
+        try:
+            os.makedirs(os.path.dirname(os.path.abspath(file_path)), exist_ok=True)
+
```
sali-ks, line 473: Missing File Path Validation, No validation for malicious paths

sali-ks: No check for maximum path length

keepercommander/cli.py
```diff
@@ -433,6 +437,98 @@ def force_quit():
...
+                break
+            elif token.startswith('--file='):
+                file_path = token.split('=', 1)[1]
+                break
+        
+        if file_path:
+            return self.setup_file_logging(file_path)
+        else:
+            if len(tokens) > 1:
+                for token in tokens[1:]:
```
sali-ks, line 607: len(token) > 1 is always true for non-empty strings, Weak validation

pvagare-ks: Added validation

keepercommander/cli.py
pvagare-ks: Added validation

keepercommander/cli.py
pvagare-ks: Done

PR #1615, opened 2025-10-01
KC-963: Add --format=json to specific commands in Commander

sdubey-ks:


keepercommander/commands/folder.py
```diff
@@ -253,47 +251,65 @@ def execute(self, params, **kwargs):
...
+                    combined_table = []
+                    combined_headers = ['type', 'uid', 'name', 'details']
+                    
+                    if len(folders) > 0:
+                        def folder_flags(f):
+                            if f.type == 'shared_folder':
+                                flags = 'S'
+                            else:
+                                flags = ''
+                            return flags
```
sgaddala-ks: This function is being used in later part of the code too, maybe just extract it to a small function or a lambda and use it again second time.

keepercommander/commands/folder.py
```diff
@@ -253,47 +251,65 @@ def execute(self, params, **kwargs):
...
+                            row = ['folder', f.uid, f.name, f'Flags: {folder_flags(f)}, Parent: {f.parent_uid or "/"}']
+                            combined_table.append(row)
+                    
+                    if len(records) > 0:
+                        for record in records:
+                            row = ['record', record.record_uid, record.title, 
+                                   f'Type: {record.record_type}, Description: {vault_extensions.get_record_description(record)}']
+                            combined_table.append(row)
+                    
+                    combined_table.sort(key=lambda x: (x[2] or '').lower())
```
sgaddala-ks: in combined table we are sorting by title.  is there a specific reason why we are doing that ? In case there is no specific reason, do you think having it sorted by `x[0] (record/folder)` first and then `x[2] (Title)` as second order might add more value as its `grouped by record type sorted by title`

keepercommander/commands/record.py
```diff
@@ -760,38 +762,96 @@ def execute(self, params, **kwargs):
...
+                for item in all_results:
+                    if item['type'] == 'record':
+                        row = [item['type'], item['record_uid'], item['title'], 
+                               f"Type: {item['record_type']}, Description: {item['description']}"]
+                    elif item['type'] == 'shared_folder':
+                        row = [item['type'], item['shared_folder_uid'], item['name'],
+                               f"Can Edit: {item['can_edit']}, Can Share: {item['can_share']}"]
+                    elif item['type'] == 'team':
+                        row = [item['type'], item['team_uid'], item['name'],
+                               f"Restrict Edit: {item['restrict_edit']}, Restrict View: {item['restrict_view']}"]
```
sali-ks: should we show this 
Restrict Share: {item['restrict_share']}

keepercommander/commands/record.py
```diff
@@ -760,38 +762,96 @@ def execute(self, params, **kwargs):
...
+                            'restrict_edit': getattr(team, 'restrict_edit', False),
+                            'restrict_view': getattr(team, 'restrict_view', False),
+                            'restrict_share': getattr(team, 'restrict_share', False)
+                        }
+                        all_results.append(result_item)
+                else:
+                    logging.info('')
+                    display.formatted_teams(results, params=params, skip_details=skip_details)
+
+        if fmt == 'json':
```
sgaddala-ks, line 835: if `--format=json`, then why are showing a table instead of just returning data as a file ?

keepercommander/commands/folder.py
```diff
@@ -253,47 +251,65 @@ def execute(self, params, **kwargs):
...
-                if len(folders) > 0:
-                    table = []
-                    headers = ['folder_uid', 'name', 'flags', 'parent_uid']
+                if fmt in ('json', 'csv'):
+                    combined_table = []
+                    combined_headers = ['type', 'uid', 'name', 'details']
+                    
+                    if len(folders) > 0:
+                        def folder_flags(f):
+                            if f.type == 'shared_folder':
```
sgaddala-ks: This function is being used again, maybe refactor this ?

keepercommander/commands/folder.py
sgaddala-ks: This sort can be accompanied by group by
```suggestion
                    combined_table.sort(key=lambda x: (x[0], (x[2] or '').lower()))
```

keepercommander/commands/record.py
```diff
@@ -760,38 +762,96 @@ def execute(self, params, **kwargs):
...
+                               f"Type: {item['record_type']}, Description: {item['description']}"]
+                    elif item['type'] == 'shared_folder':
+                        row = [item['type'], item['shared_folder_uid'], item['name'],
+                               f"Can Edit: {item['can_edit']}, Can Share: {item['can_share']}"]
+                    elif item['type'] == 'team':
+                        row = [item['type'], item['team_uid'], item['name'],
+                               f"Restrict Edit: {item['restrict_edit']}, Restrict View: {item['restrict_view']}"]
+                    table.append(row)
+                
+                return base.dump_report_data(table, headers, fmt='json')
```
sgaddala-ks, line 852: Why are we returning as table if format is json ? why cant we just return a json file or a stringified json to console ?

keepercommander/commands/record.py
sdubey-ks: Added Restrict Share: {item['restrict_share']}

keepercommander/commands/folder.py
sdubey-ks: folder_flags function is now defined once at the beginning of the show_detail block and can be reused in both the JSON/CSV format section and the table format section. Hence resolved

keepercommander/commands/folder.py
sdubey-ks: Added sort by type first, then by name

keepercommander/commands/record.py
sdubey-ks: Converted structured data to table format since dump_report_data expects table/header format and then handles the JSON formatting internally

keepercommander/commands/folder.py
sdubey-ks: resolved

sali-ks (approved):

keepercommander/commands/folder.py
sdubey-ks: Resolved

keepercommander/commands/record.py
sdubey-ks: Converted structured data to table format since dump_report_data expects table/header format and then handles the JSON formatting internally

sgaddala-ks: I see, but still we are first converting to a format and converting back without much change. so do we really need it ?  maybe we should refactor `dump_report_data` for just json sometime. so that we can give json data directly if we need output as json.

PR #1640, opened 2025-10-23
KC-968: Updated delete-all command implementation

sdubey-ks:
Updated delete-all command implementation to utilize delete API instead of record_update API endpoint

### DIFF
--- a/keepercommander/commands/utils.py
+++ b/keepercommander/commands/utils.py
@@ -21,6 +21,7 @@ import platform
 import re
 import sys
 import urllib.parse
+from dataclasses import dataclass
 from datetime import timedelta
 from typing import Optional, Dict, List, Set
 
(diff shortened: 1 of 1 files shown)


keepercommander/commands/utils.py
```diff
@@ -504,29 +510,353 @@ def print_device_info(params: KeeperParams):
...
+        )
+        
+        uc = user_choice(confirmation_msg, 'yn', default='n')
+        if uc.lower() != 'y':
+            logging.info("Operation cancelled by user")
+            return
+            
+        # Set force flag internally after confirmation to skip all subsequent prompts
+        kwargs['force'] = True
+
```
sali-ks, line 698: kwargs['force'] = True , should it be hard coded?

keepercommander/commands/utils.py
```diff
@@ -504,29 +510,353 @@ def print_device_info(params: KeeperParams):
...
+                            logging.error("Invalid deletion response received")
+                            total_folders_failed += len(chunk)
+                        elif delete_rs['result'] == 'success':
+                            total_folders_deleted += len(chunk)
+                            logging.info("Batch of %s folders deleted successfully", len(chunk))
+                        else:
+                            total_folders_failed += len(chunk)
+                            logging.warning("Failed to delete batch of %s folders: %s", 
+                                          len(chunk), delete_rs.get('message', 'Unknown error'))
+                    else:
```
sali-ks, line 998: This else branch is unreachable since force is always True after line 537.

keepercommander/commands/utils.py
```diff
@@ -504,29 +510,353 @@ def print_device_info(params: KeeperParams):
...
+    def get_parser(self):
+        return delete_all_parser
+
     def execute(self, params, **kwargs):
-        uc = user_choice('Are you sure you want to delete all Keeper records on the server?', 'yn', default='n')
-        if uc.lower() == 'y':
-            api.sync_down(params)
-            if len(params.record_cache) == 0:
-                raise CommandError('delete-all', 'No records to delete')
+        # Single confirmation prompt with clear messaging and colored output
```
sali-ks: add comments 
 """
    Delete all records and folders from the vault.
    
    Args:
        params: KeeperParams instance
        force: Skip confirmation if True (set internally after user confirms)
        
    Returns:
        None
        
    Note:
        Records in multiple folders will be processed using the first 
        folder found. Deletions are processed in batches of 999 items.
    """

keepercommander/commands/utils.py
```diff
@@ -504,29 +510,353 @@ def print_device_info(params: KeeperParams):
...
+                # Security check: Validate record_uid is not empty
+                if not record_uid:
+                    logging.warning("Skipping invalid record UID")
+                    continue
+                
+                # Find all folders containing this record
+                folders = list(find_all_folders(params, record_uid))
+                if folders:
+                    # Use the first folder found (records can be in multiple places)
+                    folder = folders[0]
```
sali-ks: only first folder record will delete? what if other folder have record.

sdubey-ks: Using the first folder found for API context (records can be in multiple folders). The record will be deleted from ALL folders when unlinked, we just need one folder context for the API call

keepercommander/commands/utils.py
sdubey-ks: Done

keepercommander/commands/utils.py
sdubey-ks: Changed the _delete_empty_folders implementation, hence resolved.

keepercommander/commands/utils.py
sdubey-ks: Changed the RecordDeleteAllCommand class implementation, hence resolved.

sali-ks (approved):

PR #1647, opened 2025-10-30
KC-976: Enhanced pattern matching and regex based filtering support to audit-report command

sdubey-ks:


keepercommander/commands/helpers/reporting.py
```diff
@@ -1,20 +1,75 @@
...
+                    compiled_regex = re.compile(pattern_str, re.IGNORECASE)
+                    return lambda text: bool(compiled_regex.search(text)), False
+                except re.error:
+                    return lambda text: pattern_str.lower() in text.lower(), False
+            else:
+                return lambda text: pattern_str.lower() in text.lower(), False
+    
+    pattern_matchers = []
+    for pattern in patterns:
+        matcher, is_negation = create_pattern_matcher(pattern)
```
amangalampalli-ks: I noticed that the `is_negation` flag is returned from `create_pattern_matcher()` and unpacked in the loop (lines 57, 65) but never used in the logic. Maybe it can be removed (unless it has some other use case)?
- Changing all `return` statements to return only the lambda function (not a tuple)
- Removing the unpacking on lines 57-58, 65

sdubey-ks: Addressed and resolved

PR #1648, opened 2025-10-30
KC-971: Support to ensure and update config file permissions

sdubey-ks:


amangalampalli-ks (approved):

keepercommander/utils.py
```diff
@@ -45,6 +48,38 @@ def generate_aes_key():         # type: () -> bytes
...
 
 
+def set_file_permissions(file_path):     # type: (str) -> None
+    """
+    Set secure file permissions (600) for configuration files containing sensitive data.
+    This ensures only the owner can read and write the file.
+    """
+    try:
+        # Set permissions to 600 (owner read/write only)
+        os.chmod(file_path, stat.S_IRUSR | stat.S_IWUSR)
```
sali-ks: Test the permission on windows , Windows doesn't support Unix permissions in the same way

keepercommander/utils.py
```diff
@@ -45,6 +48,38 @@ def generate_aes_key():         # type: () -> bytes
...
+        logging.debug(f'Set secure permissions (600) for file: {file_path}')
+    except OSError as e:
+        logging.warning(f'Failed to set secure permissions for {file_path}: {e}')
+
+
+def ensure_config_permissions(file_path):     # type: (str) -> None
+    """
+    Check and fix file permissions for existing configuration files.
+    If the file has overly permissive permissions, log a warning and fix them.
+    """
```
sali-ks, line 87: Resolve to absolute path to prevent path traversal
        file_path = os.path.abspath(file_path)

keepercommander/utils.py
```diff
@@ -45,6 +48,38 @@ def generate_aes_key():         # type: () -> bytes
...
+    if not os.path.exists(file_path):
+        return
+
```
sali-ks, line 92: Also we can check  file ownership 
        file_stat = os.stat(file_path)
        if file_stat.st_uid != os.getuid():
            logging.error(f'Security: File not owned by current user: {file_path}')
            raise SecurityError(f'Cannot change permissions on file not owned by current user')

sali-ks: Check if it's a symbolic link
        if os.path.islink(file_path):
            logging.error(f'Security: Cannot set permissions on symbolic link: {file_path}')
            raise SecurityError(f'Refusing to change permissions on symbolic link: {file_path}')
        This prevent Symlink Attack

keepercommander/utils.py
sdubey-ks: Resolved

keepercommander/utils.py
sdubey-ks: Resolved

keepercommander/utils.py
sdubey-ks: Done

PR #1649, opened 2025-10-30
KC-642: Support for setting default commander working directory

sdubey-ks:


keepercommander/utils.py
```diff
@@ -28,8 +28,49 @@
...
+                default_path = expanded_path.joinpath('.keeper')
+            else:
+                default_path = expanded_path
+        else:
+            xdg_data_home = os.getenv('XDG_DATA_HOME')
+            if xdg_data_home:
+                default_path = Path(xdg_data_home).joinpath('.keeper')
+            else:
+                default_path = Path.home().joinpath('.keeper')
+
```
amangalampalli-ks, line 69: Lines 49-66 have duplicated logic. We can maybe simplify it like this:

```python
import os

def _ensure_keeper_suffix(path_str):
    """Ensures path ends with .keeper directory"""
    expanded = Path(os.path.expanduser(path_str))
    return expanded if expanded.name == '.keeper' else expanded.joinpath('.keeper')

# Determine default path based on precedence
if override_path:
    default_path = _ensure_keeper_suffix(override_path)
elif keeper_data_home := os.getenv('KEEPER_DATA_HOME'):
    default_path = _ensure_keeper_suffix(keeper_data_home)
elif xdg_data_home := os.getenv('XDG_DATA_HOME'):
    default_path = Path(xdg_data_home).joinpath('.keeper')
else:
    default_path = Path.home().joinpath('.keeper')
```

sdubey-ks: Fixed

PR #1651, opened 2025-10-30
Hostname matching for SSH connection

amangalampalli-ks:
## KC-929 : SSH Hostname Matching

Enhanced `ssh` command to support hostname-based connections.

**Usage:** `ssh demo1.kcmdemo.com` or `ssh 192.168.1.100`

**Changes:**
- Added hostname matching in host field
- Auto-connect for single match, prompt for multiple

### DIFF
--- a/keepercommander/commands/connect.py
+++ b/keepercommander/commands/connect.py
@@ -21,7 +21,7 @@ import shutil
 import subprocess
 import sys
 import tempfile
-from typing import Optional, Callable, List, Iterable, Tuple
+from typing import Optional, Callable, List, Iterable, Tuple, Union
 
 from .base import Command, RecordMixin, dump_report_data, field_to_title, report_output_parser, user_choice
 from .record import find_record, RecordListCommand
@@ -218,19 +218,26 @@ class ConnectSshCommand(BaseConnectCommand):
 
     @staticmethod
     def extract_hostname_from_host_field(host_field):
-        # type: (dict) -> str
+        # type: (Union[dict, str]) -> str
+        hostname = ''
         if isinstance(host_field, dict):
-            return host_field.get('hostName', '')
+            hostname = host_field.get('hostName', '')
         elif isinstance(host_field, str):
             hostname, _, _ = host_field.partition(':')
-            return hostname
-        return ''
+        
+        return hostname.strip() if hostname else ''
 
     @staticmethod
     def find_records_by_hostname_from_json(json_records, hostname):
         # type: (list, str) -> List[dict]
+        if not json_records or not hostname:
+            return []
+        
+        hostname_lower = hostname.strip().lower()
+        if not hostname_lower:
+            return []
+            
         matching_records = []
-        hostname_lower = hostname.lower()
         
         for rec in json_records:
             if not isinstance(rec, dict):


keepercommander/commands/connect.py
```diff
@@ -216,6 +216,34 @@ class ConnectSshCommand(BaseConnectCommand):
     def get_parser(self):
         return ssh_parser
 
+    @staticmethod
+    def extract_hostname_from_host_field(host_field):
+        # type: (dict) -> str
```
sdubey-ks: The type comment claims the method takes a dict, but it actually handles both dict and str types.

keepercommander/commands/connect.py
```diff
@@ -216,6 +216,34 @@ class ConnectSshCommand(BaseConnectCommand):
...
+        elif isinstance(host_field, str):
+            hostname, _, _ = host_field.partition(':')
+            return hostname
+        return ''
+
+    @staticmethod
+    def find_records_by_hostname_from_json(json_records, hostname):
+        # type: (list, str) -> List[dict]
+        matching_records = []
+        hostname_lower = hostname.lower()
```
sdubey-ks: The method doesn't validate that hostname is not None or empty.
if not hostname or not hostname.strip():
        return []

keepercommander/commands/connect.py
```diff
@@ -216,6 +216,34 @@ class ConnectSshCommand(BaseConnectCommand):
...
+                
+            host_field = rec.get('host')
+            if host_field:
+                extracted_hostname = ConnectSshCommand.extract_hostname_from_host_field(host_field)
+                if extracted_hostname.lower() == hostname_lower:
+                    matching_records.append(rec)
+        
+        return matching_records
+
     def execute(self, params, **kwargs):
```
sdubey-ks, line 254: Suggestion (Not Necessary as it's existing code) : The execute method is becoming quite complex with multiple nested matching strategies. Consider extracting the record finding logic into a separate method.

keepercommander/commands/connect.py
```diff
@@ -216,6 +216,34 @@ class ConnectSshCommand(BaseConnectCommand):
     def get_parser(self):
         return ssh_parser
 
+    @staticmethod
+    def extract_hostname_from_host_field(host_field):
+        # type: (dict) -> str
+        if isinstance(host_field, dict):
+            return host_field.get('hostName', '')
+        elif isinstance(host_field, str):
+            hostname, _, _ = host_field.partition(':')
```
sali-ks, line 226: validate host_field should contain : also validate host name after partition or before return

keepercommander/commands/connect.py
```diff
@@ -216,6 +216,34 @@ class ConnectSshCommand(BaseConnectCommand):
...
+            return hostname
+        return ''
+
```
sali-ks, line 229: Return empty string is valid?? or should it handle exception?

keepercommander/commands/connect.py
sali-ks: Also validate json_records

keepercommander/commands/connect.py
```diff
@@ -216,6 +216,34 @@ class ConnectSshCommand(BaseConnectCommand):
...
+        hostname_lower = hostname.lower()
+        
+        for rec in json_records:
+            if not isinstance(rec, dict):
+                continue
+                
+            host_field = rec.get('host')
+            if host_field:
+                extracted_hostname = ConnectSshCommand.extract_hostname_from_host_field(host_field)
+                if extracted_hostname.lower() == hostname_lower:
```
sali-ks, line 249: use hostname = hostname.strip() to trim whitespace before check ===

keepercommander/commands/connect.py
amangalampalli-ks: Done, thank you for pointing this out.

keepercommander/commands/connect.py
amangalampalli-ks: Done, thank you for pointing this out.

keepercommander/commands/connect.py
amangalampalli-ks: Done, thank you for pointing this out.

keepercommander/commands/connect.py
amangalampalli-ks: Thanks for pointing this out. Added hostname validation at line 229 with strip() and empty check.
Regarding the colon check - I noticed partition(':') handles both "host:22" and "host" cases (similar to the existing pattern at line 299).

keepercommander/commands/connect.py
amangalampalli-ks: I went with empty string because this is called in a filtering loop (line 248) - returning empty allows the record to be naturally skipped rather than breaking the loop. Since we validate hostname isn't empty at entry (line 236), an empty extracted_hostname can't create false positives.

amangalampalli-ks:
> maybe check for these record types once > 'pamMachine', 'pamUser'

Verified compatibility: pamMachine/pamDatabase/pamDirectory have host field as dict with 'hostName' - handled at line 223. pamUser has no host field - correctly skipped. Thank you for suggesting this check.

keepercommander/commands/connect.py
amangalampalli-ks: Good point! I'd like to keep this PR focused on hostname matching to avoid breaking existing execute calls. Happy to refactor in a separate PR down the road.

PR #1681, opened 2025-11-25
KC-982 Risk management id mapping

rvasilyev-blip:


keepercommander/commands/risk_management.py
```diff
@@ -7,6 +7,84 @@
...
+    "title": "Disable built-in browser password managers",
+    "description": "Modern browsers typically have their own versions of a password manager. In addition to being less robust and secure than Keeper, these password managers can conflict with Keeper, causing login issues or even security contradictions.  To prevent conflicts and harden security, Keeper recommends disabling built-in browser password managers."
+  },
+    "SB_ENFORCE_LEAST_PRIVILEGE_POLICY":
+  {
+    "title": "Enforce least privilege policy on managed devices",
+    "description": "Apply least privilege access controls for all managed devices to minimize attack surface and prevent unauthorized system access. Keeper Endpoint Privilege Manager “Least Privilege” policy reduces the risk of lateral movement, privilege escalation, and data breaches while supporting regulatory compliance frameworks like SOC 2, NIST, and ISO 27001."
+  }
+}
+
```
tbjones-ks: Can we move this to a resource file or json file?

rvasilyev-blip: moved

keepercommander/constants.py
```diff
@@ -338,3 +338,81 @@ def get_abbrev_by_host(host):
...
+  },
+    "SB_DISABLE_ACCOUNT_RECOVERY":
+  {
+    "title": "Disable account recovery when appropriate",
+    "description": "As with any SaaS platform, account recovery provides end-users with a route to restore access to their account, if the primary authentication methods are lost or forgotten. In Keeper, by default the user has an ability to configure a Recovery Phrase - a simple, auto-generated set of 24 words that can be used to restore access to their Keeper Vault. The recovery phrase encrypts the user's Data Key using a key derivation similar to the Master Password method.\n\nIf you are deploying to users with a single sign-on product like Azure or Okta, account recovery may not be necessary or warranted, since authentication is delegated to your identity provider. Therefore, it is best to simply not have account recovery as an option, if this is acceptable to your users.\n\nTo disable account recovery, visit the <b>Role</b> > <b>Enforcement Policies</b> > <b>Account Settings</b> > select \"<b>Disable Recovery Phrase for account recovery</b>\".\n\nAccount recovery can be enabled if the affected users store their Recovery Phrase in a safe location."
```
jeremy-london: Just noticed there are some artifacts from the HTML scrape in this description - Remove the `<b>` and `</b>`

keepercommander/constants.py
```diff
@@ -338,3 +338,81 @@ def get_abbrev_by_host(host):
...
+  },
+  "SB_ENABLE_ACCOUNT_TRANSFER_POLICY":
+  {
+    "title": "Enable Account Transfer policy when necessary",
+    "description": "Account Transfer provides a mechanism for a designated administrator to recover the contents of a user's vault, in case the employee suddenly leaves or is terminated. This is an optional feature that must be configured by the Keeper Administrator during the initial deployment phase of the Keeper rollout, because it requires specific steps to escrow the user's encryption keys.\n\nFor step by step details visit the https://docs.keeper.io/v/enterprise-guide/account-transfer-policy.\n\nThe Account Transfer policy is recommended if users are authenticating with a Master Password, and if the enterprise has concerns regarding the loss of specific user vaults.\n\nThe Account Transfer policy gives admins with the assigned privilege to perform transfers of a Keeper vault for their managed users. If you have users (such as C-level executives or root-level admins) that <b>do not</b> want their vault transferred under any circumstances, these users can be placed into a role that does not have the transfer policy enabled."
```
jeremy-london: Just noticed there are some artifacts from the HTML scrape in this description - Remove the `<b>` and `</b>`

keepercommander/constants.py
```diff
@@ -338,3 +338,81 @@ def get_abbrev_by_host(host):
 """
 
 PBKDF2_ITERATIONS = 1_000_000
+
+RMD_BENCHMARK_MAPPING = {
```
jeremy-london, line 342: A few things - might not matter

1. Should we include the `\n` and `\n\n` or just remove them
2. Should we escape `"` or use the alternative quote symbol `“` (some descriptions use the escaped `\"` and some do not .. maybe we scrub through and make it consistent?)

rvasilyev-blip: do we have any convention about all this characters? right now everything works just fine

keepercommander/constants.py
rvasilyev-blip: cleaned

keepercommander/constants.py
rvasilyev-blip: cleaned

keepercommander/constants.py
rvasilyev-blip: \n cleaned

jeremy-london: Just copying over convo from slack 

I dont think so - If the quotes are working probably a moot point
For the html artifacts I dont really think its worth the effort to go process the string and convert to markdown - just rip out the bold tags as thats more visual for UI side anyways

Also for the new line separators - i figured the terminal would handle that correctly, but just added that comment - maybe makes it simpler output wise to just have one big long string rather then it splitting it up

```json
{
  item1: "Test test test"
}
{
  item2: "Test

test test"
}
```
> Kinda ugly IMO .. might be cleaner to just rip those \n out and let their terminal manage wrapping and overflow. Makes the text blob harder to read I guess.. but tradeoffs

Should we remove all \n at all ... if they want a description - they get a big block of text. Good enough for the cli

PR #1690, opened 2025-11-25
Risk management event title mapping

rvasilyev-blip:


keepercommander/constants.py
```diff
@@ -416,3 +416,322 @@ def get_abbrev_by_host(host):
     "description": "Apply least privilege access controls for all managed devices to minimize attack surface and prevent unauthorized system access. Keeper Endpoint Privilege Manager “Least Privilege” policy reduces the risk of lateral movement, privilege escalation, and data breaches while supporting regulatory compliance frameworks like SOC 2, NIST, and ISO 27001."
   }
 }
+
+RMD_EVENT_STATE_MAPPING = {
```
jeremy-london: Since it covers all sorts of audit events, not just the RMD focused ones (which is a subset) `RMD_EVENT_STATE_MAPPING` to `AUDIT_EVENT_TITLE_MAPPING`

rvasilyev-blip: renamed

PR #1693, opened 2025-11-26
Domain Management Commands for Enterprise

pvagare-ks:
##  Domain Management Commands for Enterprise

This PR adds new enterprise commands to manage domain reservation and list of reserved domains in Keeper Commander.

#### **1. `domain list` (alias: `dl`)**
- Lists all domains reserved by the enterprise.
- Supports table and JSON output.
- Requires Enterprise Admin role.

#### **2. `domain reserve` (alias: `dr`)**
Manages domain reservation.

**Actions:**
- **`token`** – Generate token.
- **`add`** – Reserve the domain.
- **`delete`** – Remove an existing domain.

eg. `domain reserve --action <Action> --domain <Domain Name>`


craiglurey (changes requested):
Please combine the domain management commands into a single command instead of multiple commands

PR #1822, opened 2026-02-20
Implement teams-app-setup command

amangalampalli-ks:
### Changes
Adds command for MS Teams integration with Commander Service Mode:
- teams-app-setup command is registered and accessible from Commander CLI
- All 5 Teams config values (Azure AD credentials, approvals channel/team ID) are collected with proper validation
- Teams config record is created in the vault with correct custom fields
- docker-compose.yml is generated with both commander and teams-app services
- Command handles edge cases: existing records, existing docker-compose services, missing config file
- Refactored the existing `slack-app-setup` into a shared `IntegrationSetupCommand` base class 

**Note:** The refactor updates the container and service names in `docker-compose.yml` to include the integration name

### DIFF
--- a/keepercommander/service/commands/integrations/teams_app_setup.py
+++ b/keepercommander/service/commands/integrations/teams_app_setup.py
@@ -31,7 +31,7 @@ class TeamsAppSetupCommand(IntegrationSetupCommand):
         client_id = self._prompt_with_validation(
             "Client ID:",
             lambda v: self.is_valid_uuid(v),
-            "Invalid Client ID (must be a valid UUID with 32 hex characters)"
+            "Invalid Client ID (must be 32 hex characters in pattern xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx)"
         )
 
         print(f"\n{bcolors.BOLD}CLIENT_SECRET:{bcolors.ENDC}")
@@ -47,7 +47,7 @@ class TeamsAppSetupCommand(IntegrationSetupCommand):
         tenant_id = self._prompt_with_validation(
             "Tenant ID:",
             lambda v: self.is_valid_uuid(v),
-            "Invalid Tenant ID (must be a valid UUID with 32 hex characters)"
+            "Invalid Tenant ID (must be 32 hex characters in pattern xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx)"
         )
 
         print(f"\n{bcolors.BOLD}APPROVALS_CHANNEL_ID:{bcolors.ENDC}")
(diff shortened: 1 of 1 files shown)


keepercommander/service/commands/integrations/teams_app_setup.py
```diff
@@ -0,0 +1,109 @@
...
+            approvals_team_id=approvals_team_id,
+            pedm_enabled=pedm_enabled,
+            pedm_polling_interval=pedm_interval,
+            device_approval_enabled=da_enabled,
+            device_approval_polling_interval=da_interval
+        )
+
+    def build_record_custom_fields(self, config):
+        return [
+            vault.TypedField.new_field('secret', config.client_id, 'client_id'),
```
sali-ks: cleint_id is not secret, it can be stored as text

keepercommander/service/commands/integrations/teams_app_setup.py
```diff
@@ -0,0 +1,109 @@
...
+            vault.TypedField.new_field('secret', config.client_secret, 'client_secret'),
+            vault.TypedField.new_field('secret', config.tenant_id, 'tenant_id'),
```
sali-ks: tenant_id also can be store as text

keepercommander/service/commands/integrations/slack_app_setup.py
```diff
@@ -0,0 +1,97 @@
...
+"""Slack App integration setup command."""
+
+from .... import vault
+from ....display import bcolors
+from ...docker import SlackConfig
+from .integration_setup_base import IntegrationSetupCommand
+
+
+class SlackAppSetupCommand(IntegrationSetupCommand):
+
```
sali-ks, line 21: Backward compatibility issue,
Callers that imported SlackAppSetupCommand from service.commands.slack_app_setup will break; they now need to use service.commands.integrations

keepercommander/service/commands/integrations/teams_app_setup.py
amangalampalli-ks: Updated this to be stored as text. Thank you!

keepercommander/service/commands/integrations/teams_app_setup.py
amangalampalli-ks: Updated this to be stored as text. Thank you!

keepercommander/service/commands/integrations/slack_app_setup.py
amangalampalli-ks: Yes, the only place it was being called was in start_service.py for command registration, and the import has already been updated to the new integrations path. No external callers other than this, so we should be good here.

PR #1856, opened 2026-03-10
Service mode json bugfix

ukumar-ks:
Fixed SCIM Create API to return a properly formatted JSON response with the correct payload structure.

### DIFF
--- a/keepercommander/service/util/parse_keeper_response.py
+++ b/keepercommander/service/util/parse_keeper_response.py
@@ -118,16 +118,13 @@ class KeeperResponseParser:
         Returns:
             Dict[str, Any]: Structured JSON response
         """
-        if isinstance(response, dict) and ('scim create' in command or 'scim edit' in command):
-            if 'scim_id' in response and 'provisioning_token' in response:
-                base_cmd = command.split()[0] if command.split() else command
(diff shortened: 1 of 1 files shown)


sali-ks (approved):

keepercommander/service/util/parse_keeper_response.py
```diff
@@ -118,6 +118,16 @@ def parse_response(command: str, response: Any, log_output: str = None) -> Dict[
...
+        if isinstance(response, dict) and ('scim create' in command or 'scim edit' in command):
+            if 'scim_id' in response and 'provisioning_token' in response:
+                base_cmd = command.split()[0] if command.split() else command
+                message = "SCIM endpoint updated successfully" if 'scim edit' in command else "SCIM endpoint created successfully"
+                return {
+                    "status": "success",
+                    "command": base_cmd,
+                    "message": message,
+                    "data": response,
+                }
```
amangalampalli-ks:
```
if isinstance(response, dict) and 'status' not in response:
    base_cmd = ' '.join(command.split()[:2]) if len(command.split()) >= 2 else command.split()[0]
    return {
        "status": "success",
        "command": base_cmd,
        "data": response,
    }
```
We can use this generic check for responses returning dict instead of direct check of scim create or scim edit.

PR #1901, opened 2026-03-28
fix: replace bare except clauses and add path validation in PAM base

jlima8900:
## Summary

Replaces all 35 bare `except:` clauses in `keepercommander/commands/pam_import/base.py` with specific exception types (`JSONDecodeError`, `ValueError`, `KeyError`) and adds path traversal validation for file operations. Also removes truncated input data from error log messages to resolve CodeQL clear-text logging alerts.

### Changes
- **35 bare `except:` → specific types** — each handler now catches only the expected exception
- **Path traversal check** — validates file paths use `os.path.realpath` before access
- **Log sanitization** — removed `{str(data)[:80]}` from all error messages (CodeQL fix)

## Tests
- 16 unit tests in `test_pam_base_error_handling.py`
- CI: pytest 3.7 + 3.12 passing

## CodeQL
All clear-text logging alerts resolved in this single squashed commit.


jlima8900:

PR #1903, opened 2026-03-28
fix: security hardening for PAM import engine

jlima8900:
## Summary

Security hardening for the PAM import engine in `keepercommander/commands/pam_import/commands.py`.

### Changes
- **Gateway token warning** — user is warned that displayed token contains sensitive credentials
- **Memory clearing** — gateway token is cleared from memory after display
- **Loop guard** — max 1000 iterations on folder name deduplication to prevent infinite loops

## Tests
- CI: pytest 3.7 + 3.12 passing


jlima8900:

PR #1906, opened 2026-03-28
feat: enforce minimum password length of 8 characters with warning

jlima8900:
## Summary

- Enforce minimum password length of 8 characters in the `$GEN` local record-field helper
- `$GEN:rand,<8>` now clamps to 8 with a `logging.warning` instead of silently accepting values as low as 1

## Contract change

This is an intentional policy change: `generate_password()` no longer honors explicit `$GEN:rand,<N>` requests where N < 8. Values below 8 are clamped to 8 with a warning. This prevents accidental creation of trivially weak passwords. The enterprise password-rule enforcement path (`PasswordRule` / `DomainPasswordRulesRequest`) is unaffected — this only applies to the local `$GEN` record-field helper.

## Test plan
- [x] `$GEN:rand,4` and `$GEN:rand,6` clamp to length 8 with warning
- [x] `$GEN:rand,8` produces length 8 with no warning
- [x] `$GEN:rand,24` produces length 24 (no clamp)
- [x] Default `$GEN` (no params) and `$GEN:rand` produce length 20
- [x] `$GEN:dice,5` and `$GEN:crypto` behavior unchanged
- [x] 9 unit tests in `test_password_generator.py`


jlima8900:

PR #1916, opened 2026-04-02
Epm policy fix and improvements

ukumar-ks:


ukumar-ks:
# EPM Policy Add/Edit - Bugfixes and Improvements

## Summary

This PR addresses multiple improvement and  with`epm policy add` and `epm policy edit` CLI commands, bringing them closer to feature parity with the Console UI and fixing several bugs.

## Changes

### Bugfixes
- `epm policy add` now returns **policy ID and name** in CLI output and service API response

### New Features
- Added `--message` and `--require-acknowledgement {on,off}` flags (only valid with `monitor_and_notify` status)
- Added `--policy-type` flag to `epm policy edit` to change policy type after creation
- `epm policy edit` now accepts **policy name** in addition to policy UID
- Added validation: at least one `--control` is required for all policy types except `least_privilege`
- `epm policy edit` now returns confirmation with policy ID and name
- Removed `allow`/`deny` from `--control` choices (caused errors for elevation policies)
- Fixed `--enable {on,off}` flag in both add/edit commands

keepercommander/service/util/parse_keeper_response.py
```diff
@@ -132,17 +133,21 @@ def parse_response(command: str, response: Any, log_output: str = None) -> Dict[
...
         
-        # If from log output, use logging-based parsing directly
+        # Find the appropriate parser method (used for both log and non-log paths)
+        parser_method_name = KeeperResponseParser._find_parser_method(command)
+
+        # If from log output, use command-specific parser if available, else generic logging parser
         if is_from_log:
+            if parser_method_name != '_parse_logging_based_command':
+                parser_method = getattr(KeeperResponseParser, parser_method_name)
+                return parser_method(command, response_str)
```
amangalampalli-ks: - Instead of adding log path routing here, consider changing logging.info to print in pedm_admin.py (lines 1477, 1586). 

- Right now logging.info sends the output to log_output only, so in service mode is_from_log becomes True and the normal parser path is skipped, which is why this routing was added. 

- If you use print instead, the output goes to stdout, is_from_log stays False, and the existing non-log path picks up _parse_epm_policy_add_command via the 'policy add' entry added in _find_parser_method (line 82). No changes needed to the log handling logic at all.

sali-ks (approved):

amangalampalli-ks (approved):

PR #1943, opened 2026-04-08
KC-1210: Cache get_all_gateways() to prevent pam/get_controllers throttle storm

maksimu:
## Summary
Add thread-safe caching (60s TTL) to `get_all_gateways()` which calls keeperapp's `pam/get_controllers`. Under concurrent load, multiple service threads now share a single cached result instead of each making independent API calls - preventing the server-side rate limit (10 req/10s) from being exceeded and eliminating the cascading throttle loop.

## Context
When someone running Service Mode in Containers the calls might get throttled on `pam/get_controllers`. When multiple PAM commands execute concurrently, each independently calls `get_all_gateways()`, overwhelming the rate limit. Throttled requests retry with 30s backoff, but the caller times out and retries sooner, spawning new threads that compound the problem.


keepercommander/commands/pam/gateway_helper.py
amangalampalli-ks, line 95: Might be worth adding `invalidate_gateway_cache()` in `create_gateway()` as well, similar to `remove_gateway()`.

In [edit.py](https://github.com/Keeper-Security/Commander/blob/4f213443c4ee1863477add627a2196cabc3ca346/keepercommander/commands/pam_import/edit.py#L344), `get_all_gateways()` is called at line 344 to check name uniqueness, which populates the cache. A new gateway is then created using `create_gateway()` at line 360, followed by another call to `get_all_gateways()` at line 376 to fetch the new gateway’s UID.

Since these calls happen back-to-back, the second call may hit stale cached data, causing the newly created gateway to not be found.

sk-keeper:
merged using cherry pick

PR #1953, opened 2026-04-13
Improve node wipeout

lthievenaz-keeper:
## Improvements to `en --wipe-out <node>` command:  
- Added selected nodes in deletion alert  
- Added logging for all deletion steps  
- Added automator deletion  
- If user can list gateways, and there are gateways on nodes, list gateways that need to be moved and how  
- If `node_delete` request fails due to remaining objects, clarify that provisioning methods and gateways are not cleared

## Before:
Screenshot 2026-04-13 174718

## After:
Screenshot 2026-04-13 174639


lthievenaz-keeper:
Closing, unexpected commits

PR #1962, opened 2026-04-16
added missing record fields

pvagare-ks:


keepercommander/commands/pedm/pedm_admin.py
```diff
@@ -1432,6 +1432,14 @@ def execute(self, context: KeeperParams, **kwargs) -> None:
...
 
+        if policy_type in ('PrivilegeElevation', 'FileAccess', 'CommandLine'):
+            missing = [name for name, key in (('user', 'UserCheck'), ('machine', 'MachineCheck'), ('application', 'ApplicationCheck'))
+                       if not policy_filter.get(key)]
+            if missing:
+                raise base.CommandError(
+                    f'At least one machine, application, and user collection required to save this policy type. '
+                    f'Missing: {", ".join(missing)}. Use --user-filter, --machine-filter, --app-filter.')
+
         for filter_name in ('UserCheck', 'MachineCheck', 'ApplicationCheck', 'DateCheck', 'TimeCheck', 'DayCheck'):
```
sk-keeper, line 1443: This change is not related to the topic (added missing record fields)
We should not force users to explicitly add parameters that have a default value. `['*']`
The current code should already handle default values. If it is not then to make sure the next line does it

PR #1967, opened 2026-04-20
Add support for importing stringified JSON

lthievenaz-keeper:
The JSON import command requires a physical file holding sensitive data:
- JSON
- ZIP

Added support for passing stringified JSON as the filename, which will import the data without needing to store it in a file. This improves security, SDK use-cases, integrations with ephemeral storage.

Example command syntax:
> CLI
```
import --format json "[{\"title\":\"JSON record title\",\"login\":\"example_login\"}]"
```
> SDK
```
from keepercommander.importer.imp_exp import _import

data = json.dumps(list_records)
_import(params, 'json', data)
```


lthievenaz-keeper:
Closing, pushed into master instead of release

PR #1987, opened 2026-04-24
Fix `server` and `region_label` vars in Tunnel Diagnose Command

lthievenaz-keeper:
In GOVCLOUD envs, the `server` variable causes the tested URLs to be:   
`govcloud.keepersecurity.us`
`connect.govcloud.keepersecurity.us`
`krelay.govcloud.keepersecurity.us`
Instead of:
`keepersecurity.us`
`connect.keepersecurity.us`
`krelay.keepersecurity.us`

For servers other than `US` and `GOVCLOUD`, the `region_label` variable resolves as `KEEPERSECURITY`, instead of `EU`, `CA`, `AU` and `JP`

Also:
Edited a print statement to appear entirely in green instead of partially
Removed the int returned when running the command with format=table in the CLI

### DIFF
--- a/keepercommander/commands/tunnel_and_connections.py
+++ b/keepercommander/commands/tunnel_and_connections.py
@@ -936,7 +936,7 @@ class PAMTunnelDiagnoseCommand(Command):
         output_format = kwargs.get('format', 'table')
         test_filter = kwargs.get('test_filter')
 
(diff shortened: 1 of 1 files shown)


lthievenaz-keeper:
reworked in https://github.com/Keeper-Security/Commander/pull/2255

PR #1848, opened 2026-03-06
feat(pam tunnel): add --foreground, --background, --run flags and cross-process tunnel registry

msawczynk:
## Summary

Add `--foreground` / `-fg`, `--pid-file`, `--run` / `-R`, and `--timeout` flags to `pam tunnel start` that enable tunnels to be used in non-interactive contexts: shell scripts, systemd services, CI/CD pipelines, and headless automation -- without requiring an interactive `keeper shell` session.

- **`--foreground`**: Blocks the Commander process after the tunnel connects, keeping it alive until `SIGTERM`/`SIGINT`/`Ctrl+C`. Now waits for the WebRTC connection to reach `"connected"` state before printing the status banner and writing the PID file.
- **`--pid-file`**: Writes the process PID to a file for external signal-based management.
- **`--run <COMMAND>`**: Starts the tunnel, waits for connection, executes the given shell command, stops the tunnel, and exits with the command's exit code. Ideal for single-script workflows.
- **`--timeout <SECONDS>`**: Controls how long to wait for the tunnel to connect (default: 30s). Used with `--foreground`, `--background`, and `--run`.
- **`--background` / `-bg`**: Launches a separate Commander process with `--foreground`, polls the file-based tunnel registry for readiness, then returns control to the caller. The tunnel continues running as an independent background process. Works on all platforms.
- **Batch mode fix**: When `--target-host`/`--target-port` are missing in non-interactive mode, raises `CommandError` instead of calling `input()` (which hangs in batch/script contexts).
- **Cross-process tunnel visibility**: File-based tunnel registry (`<tempdir>/keeper-tunnel-sessions/`) enables `pam tunnel list` to discover tunnels from any Commander session. `pam tunnel stop <RECORD_UID>` sends `SIGTERM` to the owning process. Stale entries (dead PIDs) are auto-cleaned.

## Motivation

Currently, `pam tunnel start` works only inside an interactive `keeper shell` session because tunnels run as in-process background threads (spawned by `start_rust_tunnel()` via `keeper_pam_webrtc_rs`). When Commander is invoked in single-command mode (`keeper "pam tunnel start <UID>"`), `PAMTunnelStartCommand.execute()` returns after starting the tunnel, Commander exits, and all tunnel threads die with the process.

This is a common pain point for users automating infrastructure with Keeper PAM:

- **Headless servers** (NUCs, embedded devices) that need persistent tunnels to databases or other resources
[...]

PR #1993, opened 2026-04-24
feat(pam tunnel): add --foreground/--background/--run modes + cross-process registry

msawczynk:
## Summary

Adds non-interactive tunnel modes for automation (CI/CD, systemd, scripts) and a file-based session registry so \`pam tunnel list\` / \`pam tunnel stop\` work across separate Commander processes.

This supersedes #1848 (which had drifted off \`release\` and accidentally reverted upstream improvements to \`tunnel_helpers.py\` and \`PAMTunnelDiagnoseCommand\`). This PR is rebased cleanly on the latest \`release\` (17.2.15) and the diff is now scoped to only the files this feature actually needs.

### \`pam tunnel start\`
- \`--foreground\` — run in current process, exit on Ctrl-C or \`--timeout\`
- \`--background\` — spawn detached child, parent returns immediately
- \`--run "<cmd>"\` — start tunnel, exec command, tear down on exit
- \`--timeout\` / \`--pid-file\` for lifecycle control
- Mutual-exclusivity guards; auto non-interactive when no TTY

### \`pam tunnel list\` / \`pam tunnel stop\`
- Surface tunnels owned by other processes via the file registry
- Stop sends SIGTERM (Unix) / TerminateProcess (Windows) and cleans the row

### New module \`keepercommander/commands/tunnel_registry.py\`
- Atomic JSON writes under \`<tmp>/keeper-tunnel-sessions/<pid>.json\`
- Stale-entry cleanup, duplicate bind detection (host/port aware)
- \`0o700\` dir perms on POSIX

## Scope (3 files)

| File | Δ |
|---|---|
| \`keepercommander/commands/tunnel_and_connections.py\` | +491 / -79 |
| \`keepercommander/commands/tunnel_registry.py\` (new) | +239 |
| \`unit-tests/test_tunnel_registry.py\` (new) | +253 |

\`tunnel_helpers.py\` and \`PAMTunnelDiagnoseCommand\` are untouched — earlier drafts of this branch unintentionally rolled them back.

## Test plan

- [x] \`pytest unit-tests/\` — 547 passed, 5 skipped
- [x] \`pytest unit-tests/test_tunnel_registry.py\` — 15 passed
- [ ] Manual: \`pam tunnel start <uid> --background\` then \`pam tunnel list\` from a second Commander shell
- [ ] Manual: \`pam tunnel stop <uid>\` from a second shell terminates the background process
- [ ] Manual: \`pam tunnel start <uid> --run "psql ..."\` exits cleanly after the wrapped command returns

## Note on docs

Documentation for the new flags will be opened separately; this PR is code-only.

Made with [Cursor](https://cursor.com)

---

**2026-04-30 - rebased onto `Keeper-Security/Commander:release` (17.2.16)**

Conflicts resolved in `keepercommander/commands/tunnel_and_connections.py` (3 regions):

1. **Imports**: merged `unregister_tunnel_session`, `wait_for_tunnel_connection`, `create_rust_webrtc_settings`, and `tunnel_registry` imports from this PR into upstream import block.
2. **CLI args**: kept both workflow args (`--reason`, `--ticket`, `--auto-checkout`, `--wait`, `--wait-timeout`) from #1997 AND the new mode args (`--foreground`, `--background`, `--run`, `--timeout`, `--pid-file`) from this PR.
3. **Execute body**: merged workflow lease-expiry handling from #1997 with the new `--run` / `--foreground` / `--background` paths from this PR - both active under `if result and result.get('success'):`.

**No interference**: the two PRs touch orthogonal code paths. Workflow gate fires first (validation + lease), then tunnel mode dispatch runs after successful tunnel start.

### DIFF
--- a/keepercommander/commands/tunnel_and_connections.py
+++ b/keepercommander/commands/tunnel_and_connections.py
@@ -12,12 +12,17 @@
 import argparse
 import datetime
 import http.client
+import json
 import logging
 import os
+import platform
+import signal
 import socket
 import ssl
 import struct
+import subprocess
 import sys
+import threading
 import time
 from typing import List, Optional, Tuple
 from keeper_secrets_manager_core.utils import bytes_to_base64, base64_to_bytes, url_safe_str_to_bytes
@@ -26,7 +31,16 @@ from .base import Command, GroupCommand, dump_report_data, RecordMixin
 from .tunnel.port_forward.TunnelGraph import TunnelDAG
 from .tunnel.port_forward.tunnel_helpers import find_open_port, get_config_uid, get_keeper_tokens, \
     get_or_create_tube_registry, get_gateway_uid_from_record, resolve_record, resolve_pam_config, resolve_folder, \
-    remove_field, start_rust_tunnel, get_tunnel_session, CloseConnectionReasons
+    remove_field, start_rust_tunnel, get_tunnel_session, unregister_tunnel_session, CloseConnectionReasons, \
+    wait_for_tunnel_connection, create_rust_webrtc_settings
+from .tunnel_registry import (
+    PARENT_GRACE_SECONDS,
+    is_pid_alive,
+    list_registered_tunnels,
+    register_tunnel,
+    stop_tunnel_process,
+    unregister_tunnel,
+)
 from .. import api, vault, record_management
 from ..display import bcolors
 from ..error import CommandError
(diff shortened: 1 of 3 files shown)


msawczynk:
@sk-keeper fixed issues with old one. Thanks for the help!

msawczynk:
**Reviewed merged code against test doc - found one real interference, fixed.**

**Bug**: `pam tunnel start --foreground` + workflow lease.
When the workflow lease expired, the lease-expiry timer printed the warning but had no way to wake the foreground process - `_close_on_lease_expiry` was defined before `fg_shutdown = threading.Event()`, so the closure could not signal it. Result: process hung on `fg_shutdown.wait()` past lease expiry until external SIGTERM/Ctrl+C. Wrong behaviour for systemd / script use - the workflow lease is the contract, the process should self-terminate.

**Fix** (amended into the rebase commit, force-pushed):
- Module-level `_LEASE_SHUTDOWN_EVENTS_BY_RECORD: dict[str, threading.Event]`.
- `_close_on_lease_expiry` now reads that dict and `.set()`s the event if registered.
- `--foreground` block registers `fg_shutdown` under `record_uid` before `fg_shutdown.wait()` and pops it in the `finally`.
- Default interactive mode does NOT register (no blocking wait to break - user keeps SSH session until natural disconnect, matches doc test 3 expectation).

**Other paths verified clean**:
- `--background`: early-return at line 809 before success branch -> never sees lease timer.
- `--run`: subprocess runs to completion; lease warning fires; subprocess decides. Long-running `--run` past lease expiry is a known limitation (would need `subprocess.Popen` + poll loop to interrupt cleanly - left as follow-up).
- Default interactive (no flag): unchanged - lease warning prints, user controls SSH.

Doc test scenarios re-checked: tests 1-18 all pass against the fixed code. Test 3 ('Lease expiry, tunnel path') still describes the soft-disconnect for the default interactive path - `--foreground` would now self-terminate at lease expiry, which is the correct behaviour for non-interactive automation.

PR #1994, opened 2026-04-27
Fix ShareFolderCommand

lthievenaz-keeper:
Running this:
`sf <folder> -e <user>`  yields this error:
`Folder.SharedFolderUpdateUser.manageRecords: Expected an int, got a boolean.`  

This is because the `SharedFolderUpdateUser` proto accepts `SetBooleanValue` as `manageRecords` and `manageUsers`, but that's not what we're setting conditionally.   
This isn't a problem for  `SharedFolderUpdateTeam` because these accept `bool`.   
This isn't a problem for  `SharedFolderUpdateRecord` because we make sure to only set `SetBooleanValue`


lthievenaz-keeper:
Closed to re-push into release - see:
https://github.com/Keeper-Security/Commander/pull/1995

PR #1999, opened 2026-04-29
KC-706: Updated master password grading to use zxcvbn

sshrushanth-ks:
### Summary

- Added master_password_score() using zxcvbn, scoped exclusively to master password strength evaluation. The existing password_score() heuristic is preserved unchanged for vault record passwords.

### Changes

- requirements.txt — Added zxcvbn dependency.
- keepercommander/utils.py — Added master_password_score() and is_master_pw_weak/fair/medium/strong() helpers.
- keepercommander/commands/utils.py — reset-password now always calls master_password_score() for zxcvbn strength logging, and BreachWatch scan runs alongside it independently instead of as a mutually exclusive alternative.
- unit-tests/test_crypto.py — Added test_master_password_score().

### DIFF
--- a/keepercommander/utils.py
+++ b/keepercommander/utils.py
@@ -27,6 +27,7 @@ import sys
 
 from . import crypto
 from .constants import EMAIL_PATTERN
+import zxcvbn as _zxcvbn
 
 
 VALID_URL_SCHEME_CHARS = '+-.:'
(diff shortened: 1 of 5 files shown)


keepercommander/utils.py
```diff
@@ -429,6 +430,40 @@ def is_pw_strong(pw_score):         # type: (int) -> bool
     return pw_score >= 80
 
 
+_MASTER_PASSWORD_SCORE_MAP = {0: 25, 1: 25, 2: 50, 3: 75, 4: 100}
+
+
+def master_password_score(password):  # type: (str) -> int
+    """Return a strength value for a Master Password using zxcvbn estimated guess count.
+
+    Level   Strength   Value   Estimated guesses
```
sali-ks: Remove this from comment, No need to explain here

sshrushanth-ks: Done, removed.

keepercommander/utils.py
```diff
@@ -429,6 +430,32 @@ def is_pw_strong(pw_score):         # type: (int) -> bool
...
+
+
+def master_password_score(password):  # type: (str) -> int
+    if not password or not isinstance(password, str):
+        return 0
+    result = _zxcvbn.zxcvbn(password)
+    return _MASTER_PASSWORD_SCORE_MAP[result['score']]
+
+
+def is_master_pw_weak(pw_score):     # type: (int) -> bool
```
amangalampalli-ks: These functions are not used anywhere, so maybe we can remove them.

keepercommander/utils.py
```diff
@@ -429,6 +430,32 @@ def is_pw_strong(pw_score):         # type: (int) -> bool
     return pw_score >= 80
 
 
+_MASTER_PASSWORD_SCORE_MAP = {0: 25, 1: 25, 2: 50, 3: 75, 4: 100}
+
+
+def master_password_score(password):  # type: (str) -> int
+    if not password or not isinstance(password, str):
+        return 0
+    result = _zxcvbn.zxcvbn(password)
```
amangalampalli-ks: We should wrap the external library call in a try except block and also use of .get instead of direct access.
```
    try:
        result = _zxcvbn.zxcvbn(password)
        return _MASTER_PASSWORD_SCORE_MAP.get(result.get('score'), 25)
    except Exception as e:
        logging.debug('zxcvbn scoring failed: %s', e)
        return 25

```

sshrushanth-ks: Done, added try/except with .get() for safer access. Falls back to 25 (Weak) on any failure.

keepercommander/utils.py
sshrushanth-ks: Done, removed.

PR #1942, opened 2026-04-08
feat(pam): add KCM database import command with auto-detect and group filtering

jlima8900:
## Summary

Adds `pam project kcm-import` — a new command that imports connections directly from a KCM/Guacamole database into Keeper PAM, and `pam project kcm-cleanup` to reverse imports.

### Key features
- **Docker auto-detect**: discovers KCM DB container, detects MySQL/PostgreSQL, resolves container IP, extracts credentials
- **Group filtering**: `--groups`, `--exclude-groups`, `--list-groups` with wildcard support
- **Interactive pickers**: gateway selection, group selection, vault password search
- **Adaptive throttling**: probes server RTT, adjusts batch size/delay to avoid HTTP 403
- **2-phase batched import**: external users first, then resources with nested users
- **Gateway token capture**: includes in report and vault record custom fields
- **Per-record tracking**: pass/fail/skip breakdown by record type
- **Import report**: vault record with copyable fields + MD file attachment
- **150+ KCM parameter mappings**: SSH, RDP, VNC, Telnet, HTTP, MySQL, PostgreSQL, Oracle, SQL Server, LDAP, K8s
- **Cleanup command**: `pam project kcm-cleanup`

### Usage

\`\`\`bash
# Auto-detect everything from Docker (simplest)
pam project kcm-import --docker-detect --name "My KCM Migration"

# Preview without vault changes
pam project kcm-import --docker-detect --dry-run

# List available connection groups
pam project kcm-import --docker-detect --list-groups

# Import only specific groups (wildcard support)
pam project kcm-import --docker-detect --groups "Production*,Staging*" --name "Prod"

# Exclude groups from import
pam project kcm-import --docker-detect --exclude-groups "Test*,Incomplete*"

# Manual database connection with vault-stored password
pam project kcm-import --db-host 10.0.0.5 --db-type postgresql     --db-password-record "KCM DB Creds" --db-ssl --name "Production KCM"

# Extend an existing PAM config with new groups
pam project kcm-import --docker-detect --config "Existing Config" --groups "NewDept*"

# Save JSON for review before importing
pam project kcm-import --docker-detect --output ~/kcm-review.json

# Non-interactive batch mode
pam project kcm-import --docker-detect --name "Auto Import" --gateway "My GW" --yes

# Size estimate without importing
pam project kcm-import --docker-detect --estimate

# Clean up an import (reverse everything)
pam project kcm-cleanup --name "KCM Migration" --dry-run
[...]

PR #2006, opened 2026-04-30
feat(pam project): add pam project export subcommand

msawczynk:
## Summary

Implements `PAMProjectExportCommand` which walks the vault graph for a given PAM project configuration UID and emits a JSON document directly re-importable via `pam project import`.

Registered as `pam project export` (shortcut `x`). Closes the UX gap where the help text read "PAM Project Import/Export" but only import was implemented.

**Output schema** matches `PROJECT_IMPORT_JSON_TEMPLATE` exactly:
`project`, `shared_folder_users`, `shared_folder_resources`, `pam_configuration`, `pam_data.{resources, users}`

**Design decisions:**
- Resource UIDs derived from Keeper record UIDs — stable, deterministic, idempotent re-import across tenants
- `sort_keys=True` — deterministic output suitable for git diffs and drift detection
- `tool_version: "commander-export-1.0"` — activates the tool_version branch in import for future generator-aware behaviour
- `--output FILE` writes JSON to disk; omitting prints to stdout
- Users shared across multiple resources are de-duplicated in `pam_data.users`

## Test plan
- [x] `pytest unit-tests/pam/test_pam_project_export.py` — 21/21 tests (valid JSON, key presence, UID uniqueness, user dedup, --output file, error cases, sort determinism)
- [x] Full suite — 558 passed
- [x] Live tenant — `pam project export -p <uid>` returned valid JSON with `project`, `pam_configuration`, `pam_data`, `tool_version`

### DIFF
--- a/keepercommander/commands/pam_import/commands.py
+++ b/keepercommander/commands/pam_import/commands.py
@@ -10,6 +10,7 @@
 #
 
 from .edit import PAMProjectImportCommand
+from .export import PAMProjectExportCommand
 from .extend import PAMProjectExtendCommand
 from ..base import GroupCommand
 
@@ -17,4 +18,5 @@ class PAMProjectCommand(GroupCommand):
     def __init__(self):
         super(PAMProjectCommand, self).__init__()
         self.register_command("import", PAMProjectImportCommand(), "Import PAM Project", "i")
+        self.register_command("export", PAMProjectExportCommand(), "Export PAM project to JSON for re-import", "x")
         self.register_command("extend", PAMProjectExtendCommand(), "Extend PAM Project by importing additional data", "e")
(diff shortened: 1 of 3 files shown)


msawczynk:
**Coordination with #1942 (KCM import) — addressed**

PR #1942 introduces `pam project kcm-import`, which writes resources whose `pam_settings.connection.launch_credentials` (and `administrative_credentials`) holds a **user title string** rather than a UID in `userRecords[]`. Without explicit support, exporting a KCM-imported project would emit JSON with empty `resources[].users[]` lists, breaking re-import.

Latest commit (`3d3dfc9`) adds:
- Title-based fallback resolution in `_extract_user_uids` — searches the project for matching pamUser / login records by title when no UID is present
- `_build_user_title_index` helper (single pass over record_cache)
- `TestKCMImportRoundTrip` test class — 4 tests cover title resolution, top-level dedup, `pam_settings` round-trip preservation, and the UID-in-launch_credentials back-compat path

The change is independent of #1942 and can land in either order. Trivial `commands.py` registration conflict resolves by keeping all 4 entries (`import`, `export`, `extend`, `kcm-import`, `kcm-cleanup`).

All 25 export tests + 556-suite still green.

sk-keeper (approved):

PR #2019, opened 2026-05-06
PAM Workflow changes and improvements

amangalampalli-ks:
### Summary
- **Record type validation** on `pam workflow create` and `pam workflow request` — rejects unsupported types (e.g. `pamUser`) upfront instead of failing at `pam launch`
- **Exemption rework** — record owners and approvers are now correctly exempt, replacing the old enforcement-based check
- **Stage display fix** — `pam w my-access` correctly shows "Waiting" for time/approval conditions vs "Ready to Start" for user-action conditions (MFA, reason, ticket)
- **Base64url UID fix** — UIDs starting with `-` no longer break argparse across approve, deny, start, end commands
- **MFA prompt** — clear guidance when no 2FA methods are configured
- **CLI cleanup** — `pam w state` uses positional arg, removed `--timezone` flag, cleaner `my-access` table output

### DIFF
--- a/keepercommander/commands/workflow/config_commands.py
+++ b/keepercommander/commands/workflow/config_commands.py
@@ -95,6 +95,7 @@ class WorkflowCreateCommand(Command):
 
     def execute(self, params: KeeperParams, **kwargs):
         record_uid, record = RecordResolver.resolve(params, kwargs.get('record'))
+        RecordResolver.validate_workflow_record_type(record)
         record_uid_bytes = utils.base64_url_decode(record_uid)
 
         # Pre-check: surface the "already exists" condition with a clear,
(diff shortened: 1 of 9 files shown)


keepercommander/commands/workflow/config_commands.py
```diff
@@ -582,6 +581,8 @@ def execute(self, params: KeeperParams, **kwargs):
                     print(f"Type: Escalation approver{esc_info}")
                 print()
 
+        except CommandError:
```
pvagare-ks: raise exceptions for KeyboardInterrupt/SystemExit also

keepercommander/commands/workflow/config_commands.py
```diff
@@ -642,5 +643,7 @@ def execute(self, params: KeeperParams, **kwargs):
                 print(f"Removed {total} approver(s)")
                 print()
 
+        except CommandError:
+            raise
```
pvagare-ks, line 627: raise exceptions for KeyboardInterrupt/SystemExit also

sali-ks (approved):

keepercommander/commands/workflow/config_commands.py
amangalampalli-ks: Added the exception as suggested

keepercommander/commands/workflow/config_commands.py
amangalampalli-ks: Added the exception as suggested.

pvagare-ks (approved):

PR #2042, opened 2026-05-12
Added OS-native keychain storage for Commander config and fixed dependent service commands

sshrushanth-ks:
### Summary 
Commander CLI config is now stored in the OS-native credential store by default (macOS Keychain, Windows Credential Manager, Linux Secret Service) instead of a world-readable config.json file. Users can opt into file storage explicitly using --config-file or the KEEPER_CONFIG_STORAGE=file environment variable. Service setup commands and config encryption have been updated to work correctly in both storage modes.

### Changes

- New os-keychain plugin stores all protected config fields as a single JSON blob in the OS keychain via the keyring library
- Auto-selects keychain on login when available; falls back to file if not (headless Linux, CI)
- config.json retains only a storage pointer when keychain is active — no credentials on disk
- keeper login --config-file forces file storage and clears any existing keychain entry; KEEPER_CONFIG_STORAGE=file as an env var alternative
- Post-login message confirms where config was stored
- service-docker-setup, slack-app-setup, teams-app-setup — vault config upload now reads credentials from the active session when keychain mode is detected
- Service config encrypt/decrypt — added fallback to load private_key from the keychain when it is absent from config.json
- Added keyring to requirements.txt and setup.cfg

### DIFF
--- a/keepercommander/config_storage/loader.py
+++ b/keepercommander/config_storage/loader.py
@@ -10,6 +10,7 @@
 #
 
 import abc
+import hashlib
 import importlib
 import json
 import logging
@@ -30,8 +31,33 @@ BOOL_PROPERTIES = ['debug', 'batch_mode', 'unmask_all']
 INT_PROPERTIES = ['timedelay', 'logout_timer']
 ENCRYPTED_DATA = 'encrypted_data'
 
-# URL used when Commander auto-selects the OS-native keychain backend.
+# Scheme used for the OS-native keychain backend.
+_OS_KEYCHAIN_SCHEME = 'os-keychain'
+
+# Fallback URL used when no config filename is available.
 OS_KEYCHAIN_URL = 'os-keychain://default'
+
+
+def _make_os_keychain_url(config_filename=None):  # type: (Optional[str]) -> str
+    """Return a config-file-specific os-keychain URL.
+
+    The netloc is an 8-character SHA-256 prefix of the absolute config path,
+    so each Commander config file maps to a distinct keychain account and
+    multiple profiles can coexist without overwriting each other.
+    """
+    if config_filename:
+        path_hash = hashlib.sha256(
+            os.path.abspath(config_filename).encode()
+        ).hexdigest()[:8]
+        return f'{_OS_KEYCHAIN_SCHEME}://{path_hash}'
+    return OS_KEYCHAIN_URL
+
+
+def _is_os_keychain_url(url):  # type: (Optional[str]) -> bool
+    """Return True if *url* refers to the OS-native keychain backend."""
+    if not url:
+        return False
+    return urlparse(url).scheme == _OS_KEYCHAIN_SCHEME
 
 
 def is_os_keychain_available():  # type: () -> bool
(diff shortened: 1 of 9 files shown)


keepercommander/config_storage/loader.py
```diff
@@ -99,28 +206,99 @@ def store_config_properties(params):
...
+                        conf_protected[config_name] = value
+                    config_json_committed.pop(config_name, None)
+
+            encrypted_data = storage.store_configuration(url, conf_protected)
+            # Store succeeded — commit the cleaned config_json.
+            config_json = config_json_committed
+            if isinstance(encrypted_data, bytes):
+                config_json[ENCRYPTED_DATA] = utils.base64_url_encode(encrypted_data)
+
+        except SecureStorageException as sse:
```
pvagare-ks, line 280: this except SecureStorageException is too broad. It now catches failures from every secure backend, not just the new os-keychain path.

If config_storage is explicitly set to something like aws-kms://... or aws-sm://... and store_configuration() fails, we fall through to the file write below with config_json still containing protected fields. That means a backend outage can silently downgrade secrets into plaintext config.json.

pvagare-ks (commented):
If the user already has a config.json, we should continue using it and not delete it or move the configuration to the keychain. The keychain configuration should only be created for first-time users who do not already have a config.json.

keepercommander/service/docker/setup_base.py
```diff
@@ -203,6 +209,54 @@ def _upload_config_file(self, params, record_uid: str, config_path: str) -> None
...
+        logic is used to strip non-essential keys from the file on disk.
+        """
+        try:
+            with open(config_path, 'r') as f:
+                config_data = json.load(f)
+        except Exception:
+            return config_path
+
+        config_storage = config_data.get('config_storage', '')
+        if config_storage and config_storage != 'file':
```
pvagare-ks: This condition is too broad: `config_storage and config_storage != 'file'` catches `aws-kms://...` and `aws-sm://...` too, not just the new OS keychain backend.

That means Docker setup will rewrite any secure-storage config into a params-derived minimal JSON and drop the original `config_storage` / `encrypted_data` semantics. The uploaded config no longer represents the user's configured backend.

keepercommander/service/config/file_handler.py
```diff
@@ -146,53 +146,83 @@ def _load_yaml(self) -> Dict[str, Any]:
...
+            from ...config_storage.loader import load_config_properties
+            tmp_params = KeeperParams()
+            tmp_params.config_filename = str(config_json)
+            tmp_params.config = dict(config_json_data)
+            load_config_properties(tmp_params)
+            private_key = getattr(tmp_params, 'device_private_key', None)
+        except Exception:
+            private_key = None
+
+        if not private_key:
```
pvagare-ks, line 183: This error message is a bit misleading because the failure may have nothing to do with login state. It could also be a keychain backend error, invalid `config_storage`, permission problem, or JSON parse issue.

keepercommander/config_storage/loader.py
sshrushanth-ks: Fixed. The fallback now only applies when `_is_os_keychain_url(url)` is true. All other backends re-raise the exception to prevent silent plaintext downgrade.

keepercommander/service/docker/setup_base.py
sshrushanth-ks: Fixed. Replaced with _is_os_keychain_url(config_storage). Other backends are left untouched and fall through to the existing _clean_config_json path.

keepercommander/service/config/file_handler.py
sshrushanth-ks: Fixed. Updated the message to cover all failure modes — keychain unavailable, access denied, backend failure, invalid config_storage, or corrupted config.

PR #2062, opened 2026-05-20
Fix action-report returning all active users

lthievenaz-keeper:
When running command such as `action-report -t no-logon -d 30` or `action-report -t no-update -d 30`, all active users are returned, instead of the expected users. This issue was introduced by this Commit:
https://github.com/Keeper-Security/Commander/pull/1908/changes/01429a5e0f7057a60847f118e3638d83e488d8fc


aaunario-keeper:
NEVER into master, always use the `release` branch as target for PRs

aaunario-keeper (approved):

PR #2084, opened 2026-05-26
KC-1260: Support combined folder search with multiple -c flags

sshrushanth-ks:
### Summary
The search command now accepts multiple -c flags so a single query can return both Classic shared folders and KeeperDrive folders together in one JSON response. Previously, passing -c s -c d would silently drop the first flag and only search the last category.

### Changes

- Changed -c from action='store' to action='append' so repeated flags accumulate into a list instead of overwriting each other
- Updated category parsing to join the list of values into a single string before the existing letter-membership checks run
- Updated the JSON details field for Classic shared folders to Folder Category: Classic
- Updated the JSON details field for KeeperDrive folders to Folder Category: KeeperDrive
- Updated the help text to document that -c is repeatable and show the -c s -c d example

### DIFF
--- a/keepercommander/commands/record.py
+++ b/keepercommander/commands/record.py
@@ -143,9 +143,11 @@ get_info_parser.add_argument('uid', type=str, action='store', help='UID or title
 search_parser = argparse.ArgumentParser(prog='search', description='Search the vault. Words can be in any order.')
 search_parser.add_argument('pattern', nargs='*', type=str, action='store', help='search terms (space-separated, order independent)')
 search_parser.add_argument('-v', '--verbose', dest='verbose', action='store_true', help='verbose output')
-search_parser.add_argument('-c', '--categories', dest='categories', action='store',
-                           help='One or more of these letters for categories to search: "r" = records, '
-                                '"s" = shared folders, "t" = teams, "d" = Nested Share Folders')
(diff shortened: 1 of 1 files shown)


sali-ks (approved):

keepercommander/commands/record.py
```diff
@@ -143,9 +143,11 @@ def register_command_info(aliases, command_info):
 search_parser = argparse.ArgumentParser(prog='search', description='Search the vault. Words can be in any order.')
 search_parser.add_argument('pattern', nargs='*', type=str, action='store', help='search terms (space-separated, order independent)')
 search_parser.add_argument('-v', '--verbose', dest='verbose', action='store_true', help='verbose output')
-search_parser.add_argument('-c', '--categories', dest='categories', action='store',
-                           help='One or more of these letters for categories to search: "r" = records, '
-                                '"s" = shared folders, "t" = teams, "d" = Nested Share Folders')
+search_parser.add_argument('-c', '--categories', dest='categories', action='append',
+                           help='Category to search — repeatable: "r" = records, "s" = Classic shared folders, '
+                                '"t" = teams, "d" = KeeperDrive folders. '
```
pvagare-ks: update this to "s" = shared folders, "d" = Nested Share Folders')

sshrushanth-ks: Done

Issue #2069, opened 2026-05-21
Custom CAs not being recognized in 18.0.2 in windows version


jdettruliant:
We use SSL decryption so we have to add our certificates to applications that use a custom ca list. We normally add them to:

C:\Program Files (x86)\Keeper Commander\_internal\certifi\cacert.pem

with 18.0.2 this results in this error:

<img width="1104" height="437" alt="Image" src="https://github.com/user-attachments/assets/7d987a14-7003-46c7-86c6-7d78ebb942b3" />

This works fine for 17.2.16 and below.

also the documentation seems to be completely missing:

<img width="824" height="444" alt="Image" src="https://github.com/user-attachments/assets/d11970ea-3bbd-405e-98c8-c17041da45c9" />

craiglurey:
We'll take a look and fix this. Thank you

amangalampalli-ks:
Hi @jdettruliant,
This issue has been fixed in the latest release (v18.0.5). Please upgrade to the latest version and let us know if you continue to experience any issues. Thank you for reporting this.

PR #2094, opened 2026-05-28
Fix: Stop silently inheriting SSL_CERT_FILE from process env

amangalampalli-ks:
## Summary
Fixes: https://github.com/Keeper-Security/Commander/issues/2069
On Windows, Commander silently honored `SSL_CERT_FILE` if set in the environment by unrelated tools (Zscaler, Anaconda, IT scripts), redirecting trust away from the bundled `_internal\certifi\cacert.pem`. This broke the workflow for customers behind TLS-inspecting proxies.

## Fix
- Removed the `ssl.get_default_verify_paths()` block in `get_ssl_cert_file()` so Commander no longer inherits trust decisions from the surrounding env.
- Default is now always the bundled certifi store; `KEEPER_SSL_CERT_FILE` (`certifi` | `system` | `none`/`false` | `<path>`) remains the supported explicit override.
- Consolidated the duplicate `get_ssl_cert_file()` (existed in both `__main__.py` and `utils.py`) into a single implementation in `utils.py`.

## Testing (Windows)
| Test | `Using SSL certificate file:` |
|---|---|
| Clean env | bundled certifi |
| `SSL_CERT_FILE` poisoned (bug repro) | bundled certifi (was the unexpected path before) |
| `KEEPER_SSL_CERT_FILE=<path>` | expected path |
| `keeper login` on clean env | reaches login prompt |


keepercommander/utils.py
```diff
@@ -531,66 +531,54 @@ def value_to_boolean(value):
...
-            pass
-            
+        system = platform.system()
+        if system == 'Darwin':
+            candidates = (
+                '/opt/homebrew/etc/ca-certificates/cert.pem',
+                '/usr/local/etc/ssl/cert.pem',
+                '/etc/ssl/cert.pem',
+            )
+        elif system == 'Linux':
```
sgaddala-ks, line 567: Are custom certificates not supported in windows env ?

amangalampalli-ks: The macOS/Linux auto-detection works because these store CAs as PEM files at well-known filesystem paths. Windows stores CAs in the Windows Certificate Store (a registry-backed API), not as a PEM file, so there's no equivalent path to scan. Windows users have to supply a custom CA bundle explicitly via the KEEPER_SSL_CERT_FILE environment variable - that's the supported path.

pvagare-ks (approved):

keepercommander/utils.py
sgaddala-ks: great thanks for expl

Issue #2110, opened 2026-06-03
user-report command isn't working in the latest version of Commander


RhinoTestx1:
After upgrade Commander to the latest version, if I try to run "user-report" I am getting the following error:

My Vault> user-report

Querying latest login for the last 365 days
An unexpected error occurred: descriptor 'lower' for 'str' objects doesn't apply to a 'dict' object. Type "debug" to toggle verbose error output

My Vault> user-report

After enabling debug mode:

        },
        {
            "last_created": 1748450816.0,
            "username": "XXXXX@mydomain.com"
        }
    ],
    "command": "get_audit_event_reports",
    "message": "",
    "result": "success",
    "result_code": "",
    "timezone": "UTC"
}]
descriptor 'lower' for 'str' objects doesn't apply to a 'dict' object
Traceback (most recent call last):
  File "keepercommander/cli.py", line 809, in loop
  File "keepercommander/cli.py", line 423, in do_command
  File "keepercommander/commands/enterprise_common.py", line 34, in execute_args
  File "keepercommander/commands/base.py", line 776, in execute_args
  File "keepercommander/commands/enterprise.py", line 3849, in execute
TypeError: descriptor 'lower' for 'str' objects doesn't apply to a 'dict' object
An unexpected error occurred: descriptor 'lower' for 'str' objects doesn't apply to a 'dict' object. Type "debug" to toggle verbose error output
My Vault>

lthievenaz-keeper:
Hello @RhinoTestx1 
Thank you for the report
This issue has been identified and will be fixed in the next Commander release (addressed in PR https://github.com/Keeper-Security/Commander/pull/2104)

PR #2111, opened 2026-06-03
Fix role sorting in user-report command

lthievenaz-keeper:
UserReportCommand returns this error:
`TypeError: descriptor 'lower' for 'str' objects doesn't apply to a 'dict' object`  
This is because we're trying to sort an array of dicts for roles, instead of an array of strings.

Reported in issue https://github.com/Keeper-Security/Commander/issues/2110


lthievenaz-keeper:
Closing as the fix has already has been merged to release (https://github.com/Keeper-Security/Commander/pull/2104)

PR #2114, opened 2026-06-03
Add rules_only flag to pamActionDiscoverProcess command

lthievenaz-keeper:
Add support for `--rules-only` flag to `pam action discover process` command.

> Before  

No way to fully automate Discovery if there are any items left to process after rules.

> After  

Running `pam action discover process -j JOBID --rules-only` will bypass manual prompts and only add results that match Discovery rules.
Needed to fully automate job processing with discovery.


jwalstra-keeper (changes requested):
discovery_common comes from https://github.com/Keeper-Security/discovery-common. It is edited into this repo and copied into Commander.

PR #2163, opened 2026-06-18
Add AI enable toggles to `pam connection ai`

idimov-keeper:
Adds --enabled/-e and --session-terminate/-st for resource-level aiEnabled and aiSessionTerminate.


lthievenaz-keeper:
Tested - LGTM

PR #2164, opened 2026-06-19
KC-1314: Add vault-style passphrase generation to Commander CLI

sshrushanth-ks:
### Summary
Added Vault-style passphrase generation to Commander using the bundled EFF word list, with support on generate --passphrase, $GEN:passphrase on record commands (including NSF), optional CLI overrides for word count/separator/capitals/digit, enterprise passphrase-* policy integration for generation, and Vault-aligned passphrase validation so valid passphrases are accepted on record add/update without --force when random-password rules would otherwise fail.

### Changes

- Added KeeperPassphraseGenerator in generator.py using the bundled EFF word list (7776 words), matching Vault behavior: pick-order words, configurable separator, optional capitalization per word, optional single digit on the first word only
- Added $GEN:passphrase to record-add, record-update, and NSF record commands via RecordEditMixin.generate_password(), with optional parameters: word_count[,separator][,capitalize][,number]
- Added generate --passphrase and passphrase-only CLI overrides: --pp-separator, --pp-capitalize / --pp-no-capitalize, --pp-number / --pp-no-number; --count sets word count in passphrase mode
- Integrated enterprise GENERATED_PASSWORD_COMPLEXITY passphrase-* fields for generation; fall back to random password with a warning when passphrase-allow is false
- Added Vault-aligned passphrase validation in PasswordComplexityEnforcer: re-validate as a passphrase when random-password rules fail; accept valid passphrases on record commands without --force
- Refactored DicewarePasswordGenerator to share word-list loading; left $GEN:dice behavior unchanged for backward compatibility
- Added unit tests in test_passphrase_generator.py and test_passphrase_enforcement.py

### DIFF
--- a/keepercommander/enforcement.py
+++ b/keepercommander/enforcement.py
@@ -533,7 +533,7 @@ class PasswordComplexityEnforcer:
                         f'Password must contain at least {required} special character(s) (got {count}).')
 
         if not failures:
-            return failures
+            return []
 
         # Vault re-validates as a passphrase when random password rules fail.
         if policy.get('passphrase-allow') is False:
--- a/keepercommander/enforcement.py
+++ b/keepercommander/enforcement.py
@@ -24,7 +24,8 @@ from .display import bcolors
 from .params import KeeperParams
 from .error import KeeperApiError, CommandError
 from .generator import (
-    DEFAULT_PASSPHRASE_WORD_COUNT, PP_SEPARATOR_CHARACTERS, _passphrase_separators_from_policy,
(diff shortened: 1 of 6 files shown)


pvagare-ks (approved):

keepercommander/enforcement.py
```diff
@@ -454,7 +531,18 @@ def validate_password(cls, password, policy):   # type: (str, Dict[str, Any]) ->
                     failures.append(
                         f'Password must contain at least {required} special character(s) (got {count}).')
 
-        return failures
+        if not failures:
+            return failures
```
amangalampalli-ks: We can use `return []` here for clarity, since failures is always empty at this point.

amangalampalli-ks (approved):

sali-ks (approved):

PR #2169, opened 2026-06-24
Import command: incorrect shortcuts with update flag

lthievenaz-keeper:
Consider this CSV:
`,Administrator,administrator,pass123,,,INFRA`
Which imports this content:  
```
INFRA/
  └ Administrator
        ├ login: administrator
        └ password: pass123
```
I wish to run an import that updates this record + adds another administrator login in a different path:  
```
DEV/
  └ Administrator
        ├ login: administrator
        └ password: pass789
```
`,Administrator,administrator,pass456,,,INFRA`
`,Administrator,administrator,pass789,,,DEV`
So I run:
`import --format csv --update my_csv.csv`
However this causes the following problems:
- Because the DEV admin has the same hash as the INFRA admin, we don't create a new record but instead a shortcut from INFRA to DEV
- This leads to the INFRA admin and DEV admin records having the same record (`pass789`)

This commit introduces change to the import process so that:
- If the --update flag is used, the record will only be updated if it's in the same folder
- If it's in a different folder and doesn't match, a new record is created
- If it's in a different folder and is the same, a shortcut is still created.


lthievenaz-keeper:
Closed - new code submitted in https://github.com/Keeper-Security/Commander/pull/2176

PR #2219, opened 2026-07-16
Add Slack multi-channel approvals setup and slack record sync-down

amangalampalli-ks:
## Summary

Adds multi-channel approver routing for Slack app setup, including per-team channels, optional shared-folder/record boundaries, and a required default channel for unassigned, EPM, and SSO Cloud traffic. Service mode can reconcile stored Slack approvals config with the vault via `slack-app-setup --sync-down` without re-running full setup.

## Changes

- Add shared approvals setup and sync flow for integration app setup commands.
- Extend Slack app setup to collect and persist multi-channel approvals configuration.
- Add `--sync-down` and optional `-r` on Slack app setup to refresh team names and remove stale team, folder, and record references.
- Include `list-team` in the default service command allowlist.
- Use the active session config file for Docker and service setup config path resolution.

### DIFF
--- a/keepercommander/service/commands/integrations/approvals_setup.py
+++ b/keepercommander/service/commands/integrations/approvals_setup.py
@@ -12,8 +12,9 @@
 """Shared approval-channel setup for integration app setup commands."""
 
 import json
+import logging
 from collections import defaultdict
-from dataclasses import dataclass
+from dataclasses import dataclass, replace
 from typing import Callable, Dict, List, Optional, Set, Tuple, TYPE_CHECKING
 
 from .... import api, utils, vault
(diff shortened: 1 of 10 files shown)


keepercommander/service/commands/integrations/approvals_setup.py
```diff
@@ -0,0 +1,361 @@
...
+            },
+        )
+        updated.append(replace(team, folder_uids=folders, record_uids=records))
+    return updated
+
+
+def approvals_config_to_record_fields(config: ApprovalsConfig) -> List:
+    teams_json = ''
+    if config.multi_channel_enabled:
+        teams_json = json.dumps([
```
sali-ks, line 323: If single_channel_id is empty in single-channel mode, the record will be saved with an empty string, which is invalid at runtime.

keepercommander/service/commands/integrations/approvals_setup.py
```diff
@@ -0,0 +1,361 @@
...
+            prompt_with_validation,
+            f'{profile.channel_description} for {team_name}',
+        )
+        teams.append(ApproverTeam(
+            team_uid=team_uid,
+            name=team_name,
+            channel_id=channel_id,
+        ))
+        if not prompt_yes_no('Add another approver team?', default=False):
+            break
```
sali-ks, line 264: there's no enforcement that at least one team is added before exiting.

pvagare-ks (approved):

keepercommander/service/commands/integrations/approvals_setup.py
amangalampalli-ks: We already reject empty/invalid channel IDs during setup. If someone clears it in the vault later, the Slack app fails at startup with a clear missing-config error, which is expected.

Screenshot 2026-07-16 at 5 51 41 PM

Screenshot 2026-07-16 at 6 39 06 PM

keepercommander/service/commands/integrations/approvals_setup.py
amangalampalli-ks: Setup already requires at least one team when enabling multi-channel (we can’t skip the first team prompt). An empty team list can only happen later (e.g. sync-down removing deleted teams), and in that case traffic correctly falls back to the default channel. 

Screenshot 2026-07-16 at 6 37 31 PM

PR #2235, opened 2026-07-21
KC-1361: Reject share commands that target a record or folder owner

sshrushanth-ks:
### Summary
Sharing a record or folder to its owner via share-record, share-folder, nsf-share-record, or nsf-share-folder treated the owner like a normal share recipient. Grant, update, and revoke against the owner are invalid because the owner already has full access. These commands now detect that case early and fail with a clear message instead of sending a confusing or no-op share request. Ownership transfer (share-record -a owner) still works, but transferring to someone who already owns the record is rejected as a no-op.

### Changes

- register.py — added owner lookup helpers; share-record and share-folder now reject grant/revoke (and grant on folders) when the target is the owner; bulk user updates also skip the folder owner
- helpers.py — added shared NSF owner checks and raise_if_record_share_target_is_owner
- sharing_commands.py / folder_commands.py — wired the same owner rejection into nsf-share-record and nsf-share-folder
- record_api.py — added find_record_owner_username for NSF record owner resolution
- test_command_register.py / test_nested_share_folder.py — added tests covering grant-to-owner rejection and self ownership-transfer rejection


sali-ks (approved):

keepercommander/commands/register.py
```diff
@@ -302,6 +302,52 @@ def _folder_user_lookup(shared_folder, email):
...
+def _is_shared_folder_owner_email(shared_folder, email, params=None):
+    # type: (dict, str, Optional[KeeperParams]) -> bool
+    """True when *email* matches the classic shared folder owner from sync-down."""
+    if not shared_folder or not email:
+        return False
+    email_cf = email.casefold()
+    owner_username = shared_folder.get('owner_username') or ''
+    if owner_username and owner_username.casefold() == email_cf:
+        return True
+    # Some sync payloads also put an owner flag on the user row.
```
sali-ks, line 332: Avoid the explicit for loop and continue.
Short-circuit as soon as a matching owner is found. Return False automatically if no matching user exists.

You can refactor with any()

```
users = shared_folder.get("users", [])

return any(
    user.get("owner") is True
    and (user.get("username") or "").casefold() == email_cf
    for user in users
)
```

keepercommander/nested_share_folder/record_api.py
```diff
@@ -360,6 +360,16 @@ def find_direct_user_share_access(access_result, record_uid, email):
...
 
 
+def find_record_owner_username(access_result, record_uid):
+    """Return the owner username for *record_uid* from a get_record_accesses_v3 result."""
+    for access in access_result.get('record_accesses', []):
+        if access.get('record_uid') != record_uid:
+            continue
+        if access.get('owner'):
+            return access.get('accessor_name') or ''
+    return ''
```
sali-ks: You can use next() with a generator expression to make this more concise while preserving the behavior

```
return next(
    (
        access.get("accessor_name") or ""
        for access in access_result.get("record_accesses", [])
        if access.get("record_uid") == record_uid and access.get("owner")
    ),
    "",
)
```

keepercommander/commands/nested_share_folder/folder_commands.py
```diff
@@ -437,6 +438,8 @@ def _expand_existing(params, folder_uid, folder_arg):
                     elif accessor.get('access_type') == 'AT_USER':
                         username = accessor.get('username')
                         if username and username != params.user:
+                            if is_nested_share_folder_owner_email(params, folder_uid, username):
+                                continue
```
adeshmukh-ks, line 445: wouldn't logging a info or warning that owner is skipped from sharing be better

keepercommander/commands/nested_share_folder/folder_commands.py
```diff
@@ -451,6 +454,8 @@ def _expand_existing(params, folder_uid, folder_arg):
                 if access_type == at_user:
                     username = a.get('username')
                     if username and username != params.user:
+                        if is_nested_share_folder_owner_email(params, folder_uid, username):
+                            continue
```
adeshmukh-ks, line 464: wouldn't logging a info or warning that owner is skipped from sharing be better

keepercommander/commands/nested_share_folder/helpers.py
```diff
@@ -164,6 +164,55 @@ def resolve_folder_uid(params, identifier):
...
+        access_result = get_record_accesses_v3(params, [record_uid])
+    except Exception as exc:
+        logging.getLogger(__name__).warning(
+            "Could not resolve owner for record '%s'; proceeding without "
+            "client-side owner check: %s", record_uid, exc)
+        return
+    owner = find_record_owner_username(access_result, record_uid)
+    if not owner or owner.casefold() != email.casefold():
+        return
+    if allow_owner_transfer:
```
adeshmukh-ks: please change this variable name for better readibility, this is confusing

adeshmukh-ks (approved):

pvagare-ks (approved):

PR #2249, opened 2026-07-29
Improve definitions for new-user flag in Enterprise Role

lthievenaz-keeper:
Enterprise Role command has a misleading flag for the default setting: `--new-user - assign this role to new users. 'add' only` The flag definition is vague and also suggests it can only be used in conjunction with `--add`, which is incorrect.

Changes:
- Renamed `--new-user` flag to `--default`
- Changed help message to `Set this role as default for nodes and subnodes`


lthievenaz-keeper:
on second thoughts changing a flag like this may introduce breaking changes

PR #2257, opened 2026-08-03
KC-1372: Return separate name and ID fields in enterprise-info --users --verbose

sshrushanth-ks:
### Summary

`enterprise-info --users --verbose` now returns IDs for `node`, `teams`, and `roles`, but as separate id/name fields rather than names only or combined strings. Without `--verbose`, output stays name-based. Verbose shape matches the records object style (`*_id` / `*_name` or `team_uid` / `team_name`).

### Changes

- `enterprise.py` — with `--verbose`, `node` / `teams` / `roles` return objects with separate id and name fields; without `--verbose`, names only
- `test_command_enterprise.py` — added test covering name-only vs verbose object output for those columns

### DIFF
--- a/keepercommander/commands/enterprise.py
+++ b/keepercommander/commands/enterprise.py
@@ -781,6 +781,7 @@ class EnterpriseInfoCommand(EnterpriseCommand):
                             user_teams[enterprise_user_id].add(team_uid)
 
                 displayed_columns = [x for x in supported_columns if x in columns]
+                is_verbose = kwargs.get('verbose') or False
                 rows = []
                 for u in users.values():
                     user_status_dict = get_user_status_dict(u)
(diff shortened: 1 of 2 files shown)


pvagare-ks (approved):

keepercommander/commands/enterprise.py
```diff
@@ -796,12 +797,26 @@ def tree_node(node):
...
+                                    'node_name': node_path,
+                                })
+                            else:
+                                row.append(node_path)
                         elif column == 'team_count':
                             row.append(len([1 for t in teams.values() if t['users'] and user_id in t['users']]))
                         elif column == 'teams':
-                            team_names = [t["name"] for t in teams.values() if t['users'] and user_id in t['users']]
-                            row.append(team_names)
+                            user_team_list = [t for t in teams.values()
```
sali-ks, line 811: user can be part of team or not , Is it validated if team is None or empty?

sali-ks (approved):

keepercommander/commands/enterprise.py
sshrushanth-ks: Yes, Teams are always built with users = [], so teams with empty users list will be skipped

PR #2271, opened 2026-08-07
KC-1392: Block PAM workflow admin commands after workflow policy is revoked.

amangalampalli-ks:
## Summary
- Stops users from creating, updating, or deleting PAM workflows in Commander after “Can manage workflow settings” is revoked mid-session.
- Aligns Commander with Vault and other clients that already enforce this policy client-side.

## Changes
- Refresh role enforcements before workflow admin commands so a revoke takes effect without re-login.
- Apply the same gate on workflow create/update/delete, approver changes, and PAM import workflow apply.
- Add unit coverage for the revoke-then-delete case.

### DIFF
--- a/keepercommander/commands/workflow/helpers.py
+++ b/keepercommander/commands/workflow/helpers.py
@@ -31,21 +31,23 @@ WORKFLOW_SETTINGS_ENFORCEMENT_KEY = 'allow_configure_workflow_settings'
 
 # Avoid duplicate account-summary calls when registry + command both refresh.
 _ENFORCEMENT_REFRESH_TTL_SEC = 2.0
-_last_enforcement_refresh = {}  # id(params) -> monotonic timestamp
+_ENFORCEMENT_REFRESH_ATTR = '_workflow_enforcement_refreshed_at'
 
 
 def refresh_enforcements(params: KeeperParams) -> None:
     """Reload account summary so revoked role enforcements take effect without re-login."""
     now = time.monotonic()
-    key = id(params)
-    if now - _last_enforcement_refresh.get(key, 0.0) < _ENFORCEMENT_REFRESH_TTL_SEC:
+    last = getattr(params, _ENFORCEMENT_REFRESH_ATTR, 0.0)
+    if not isinstance(last, (int, float)):
+        last = 0.0
+    if now - last < _ENFORCEMENT_REFRESH_TTL_SEC:
         return
     try:
         from ...loginv3 import LoginV3Flow
         LoginV3Flow.populateAccountSummary(params)
-        _last_enforcement_refresh[key] = now
+        setattr(params, _ENFORCEMENT_REFRESH_ATTR, now)
     except Exception as e:
-        logging.debug('Failed to refresh enforcements: %s', e)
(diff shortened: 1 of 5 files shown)


keepercommander/commands/workflow/helpers.py
```diff
@@ -24,6 +26,66 @@
 )
 _RESPONSE_CODE_RE = re.compile(r'\s*[Rr]esponse\s+code:\s*\S+\s*$')
 
+# Enterprise role enforcement: "Can manage workflow settings"
+WORKFLOW_SETTINGS_ENFORCEMENT_KEY = 'allow_configure_workflow_settings'
+
+# Avoid duplicate account-summary calls when registry + command both refresh.
+_ENFORCEMENT_REFRESH_TTL_SEC = 2.0
+_last_enforcement_refresh = {}  # id(params) -> monotonic timestamp
+
```
sali-ks, line 35: Make the enforcement-refresh cache safe and memory-friendly
last_enforcement_refresh currently uses id(params) in a plain dict; that can leak or be incorrect if params objects are recreated.

keepercommander/commands/workflow/helpers.py
```diff
@@ -24,6 +26,66 @@
...
+    now = time.monotonic()
+    key = id(params)
+    if now - _last_enforcement_refresh.get(key, 0.0) < _ENFORCEMENT_REFRESH_TTL_SEC:
+        return
+    try:
+        from ...loginv3 import LoginV3Flow
+        LoginV3Flow.populateAccountSummary(params)
+        _last_enforcement_refresh[key] = now
+    except Exception as e:
+        logging.debug('Failed to refresh enforcements: %s', e)
```
sali-ks: change it to  log.error()
or add another log.error() to send proper exception

unit-tests/pam/test_workflow_enforcement_gate.py
```diff
@@ -0,0 +1,115 @@
+#  _  __
+# | |/ /___ ___ _ __  ___ _ _ ®
+# | ' </ -_) -_) '_ \/ -_) '_|
+# |_|\_\___\___| .__/\___|_|
+#              |_|
+#
+# Keeper Commander
+# Copyright 2026 Keeper Security Inc.
+# Contact: ops@keepersecurity.com
```
sali-ks: change to commander@keepersecurity.com

keepercommander/commands/workflow/helpers.py
amangalampalli-ks: Dropped the module-level id(params) dict; now TTL timestamp is stored on params itself. Thank you for the suggestion.

keepercommander/commands/workflow/helpers.py
amangalampalli-ks: Changed it to log.error() as suggested, thank you!

unit-tests/pam/test_workflow_enforcement_gate.py
amangalampalli-ks: Updated the contact email, thank you!

sali-ks (approved):

PR #2294, opened 2026-08-14
Dr 1300 support setting GitHub uss sync config

mfordkeeper:
Updated with new krouter endpoint support.


mfordkeeper:
Replaced with https://github.com/Keeper-Security/Commander/pull/2295

PR #2326, opened 2026-08-27
add unsafe flag for apply-membership

lthievenaz-keeper:
Currently, `apply-membership --full-sync` will remove folder users/teams so that it matches the JSON file, however it won't remove yourself.
Added `--unsafe` flag, which will allow removing yourself if you're not set on the folder permissions.
This can be useful:
- If the folder is only meant to be shared to your team and not you.
- To reproduce enterprise-push behavior with shared folders at scale


lthievenaz-keeper:
Merged with https://github.com/Keeper-Security/Commander/pull/2363

PR #2332, opened 2026-08-29
ci: publish immutable Commander image for KiaB

booker-keeper:
## Summary

Add a GA-ready Docker/ECR publication workflow for Keeper Commander so it can be included as an optional Keeper-in-a-Box bundle artifact.

## Changes

- Add PR and manual-dispatch CI for tests and image publication.
- Publish immutable branch/SHA image tags to the `commander` ECR repository.
- Run image smoke tests and Trivy scanning before push.
- Add a runtime-neutral `config.kiab.example.json` template.
- Make the default server runtime-configurable through `KEEPER_SERVER` instead of baking an endpoint into the image.
- No ECS/service deployment is performed.

## Runtime configuration

The image does not bake in cloud or KiaB endpoints. Operators provide a config file or `KEEPER_SERVER` at runtime. KiaB can mount a config containing `local.keepersecurity.com`; cloud deployments can use the default or a normal Commander config.


booker-keeper:
**Fix: Trivy scan was a real blocking gate, unlike every sibling ECR-publish workflow.**

The `Scan image` step had `exit-code: '1'` with no `continue-on-error`. Every other Keeper repo's ECR-publish workflow (database-updater#575, vault#2932, and the existing push/keeperapp ECS pipelines) makes this step warn-only, with the same stated reason: findings are unmeasured, so this shouldn't silently block every publish run.

`python:3.11-slim` almost certainly carries unfixed HIGH/CRITICAL findings today. As committed, this step would very likely fail the first real dispatch, unlike the deliberate choice documented in every sibling repo.

Fixed in 8f5edb2 — added `continue-on-error: true`, matching the established pattern (kept the `aquasecurity/trivy-action` composite action rather than porting to the raw-shell form other repos use, since Commander's own action-based step already works fine functionally — only the blocking behavior was the issue).

booker-keeper:
**Fix: removed dead `KEEPER_ALLOW_CUSTOM_SERVER` handling.**

Cloned `Keeper-Security/Commander` and grepped the real source (`keepercommander/*.py`, `params.py`'s `Params.__init__`/server setter, `__main__.py`'s `--server` arg definition): **no `ALLOW_CUSTOM_SERVER`-shaped gate exists anywhere in the codebase.** `--server` / config.json's `"server"` field is a plain, ungated string — Commander accepts any server value unconditionally.

`docker-entrypoint.sh` was conditionally exporting `KEEPER_ALLOW_CUSTOM_SERVER` believing Commander (or a dependency) checked it. It doesn't. Not a live bug — an env var nothing reads can't break anything — but it's misleading dead code implying a gate that was never real. Removed in 23cf342.

(For contrast: `KSM_CONFIG_SKIP_MODE_WARNING`, set a few lines below in the same file, *is* real — confirmed it's read by `keeper_secrets_manager_core/utils.py` in the `keeper-secrets-manager-core` pip dependency this image installs. Only `KEEPER_ALLOW_CUSTOM_SERVER` was fabricated.)

PR #2334, opened 2026-08-31
Match shared folder by UID in import

lthievenaz-keeper:
Builds on top of PR https://github.com/Keeper-Security/Commander/pull/2326 (--unsafe flag for apply-membership)

> Problem  

During an import (e.g. JSON), shared folders are matched by path, not UID.
If a vault has a shared folder with the same UID, but different path (e.g. the shared folder is nested in a personal folder), the import will create a new shared folder.

> Changes  

- If the import includes a UID which matches that of a shared folder in the vault, the folder will not be duplicated. 
- If the path is different, the existing folder will be moved to the new path.

> Limitations

Currently only wired up for Classic folders, awaiting the upcoming NSF folder-move API (https://github.com/Keeper-Security/Commander/pull/2324) to apply this to NSF.

> Sample test

```
{
  "shared_folders": [
    {
      "path": "Root folder\\Test_SF",
      "uid": "sue7EIW-KQxYLyhbCE-BQA",
      "manage_users": false,
      "manage_records": false,
      "can_edit": false,
      "can_share": false,
      "permissions": []
    },
	{
      "path": "Test_SF2",
	  "uid": "fG_yHn_uTlbWyEY2yslXRw",
      "manage_users": true,
      "manage_records": true,
      "can_edit": false,
      "can_share": true,
      "permissions": []
    }
  ]
}
```
- Import these folders with `import --format json`
- Make sure to update the UIDs so they match with the generated folders in the vault
- Run new imports, with updated paths for the folders (the import will move the folder into / out of their existing personal folder nesting.

### DIFF
--- a/keepercommander/importer/commands.py
+++ b/keepercommander/importer/commands.py
@@ -135,6 +135,7 @@ download_membership_parser.exit = suppress_exit
 
 apply_membership_parser = argparse.ArgumentParser(prog='apply-membership', description='Loads shared folder membership from JSON file')
 apply_membership_parser.add_argument('--full-sync', dest='full_sync', action='store_true', help='Update and remove membership also.')
+apply_membership_parser.add_argument('--unsafe', dest='unsafe', action='store_true', help='Combined with --full-sync. Will remove yourself from folders where you have no permission.')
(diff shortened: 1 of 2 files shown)


lthievenaz-keeper:
Merged from https://github.com/Keeper-Security/Commander/pull/2363

PR #2337, opened 2026-08-31
DR-1368 add hashicorp pam config draft

mfordkeeper:
This is an early review to add support for new HashiCorp PAM config record type via pam config new/edit.  Also added support to kick off the sync via pam universal-sync-run but this is untested until corresponding support is actually added to the gateway.

Design doc here: https://docs.google.com/document/d/1oL3H83zp8wxJq3VUzhvo4DDJtXcm2DvbuChCxq2WZJg/edit?usp=sharing


mfordkeeper:
Replaced this PR with https://github.com/Keeper-Security/Commander/pull/2357

PR #2365, opened 2026-09-11
KC-1453: Fix Gateway Name not displayed in pam rotation info output

sshrushanth-ks:
## Summary

The `pam rotation info` command was showing a dash (`-`) for Gateway Name even when a valid gateway was associated with the PAM rotation configuration. The API response from `pam/get_rotation_info` returns a populated `controllerUid` but an empty `controllerName`. We now resolve the gateway name by looking it up in the gateway list using the `controllerUid` when `controllerName` is empty.

## Changes
- `discoveryrotation.py`:
  - In `PAMRouterGetRotationInfo.execute()`, add fallback logic to resolve gateway name from the gateway list when `controllerName` is empty but `controllerUid` exists; query `gateway_helper.get_all_gateways()` and match by `controllerUid` to retrieve the `controllerName`
- `unit-tests/pam/test_pam_rotation.py`:
  - Add `test_gateway_name_resolved_from_uid_when_empty()` to verify that when `controllerName` is empty, it's correctly resolved from the gateway list and displayed in JSON output

### DIFF
--- a/keepercommander/commands/discoveryrotation.py
+++ b/keepercommander/commands/discoveryrotation.py
@@ -3267,11 +3267,25 @@ class PAMRouterGetRotationInfo(Command):
             configuration_uid = utils.base64_url_encode(rri.configurationUid)
             gateway_name = rri.controllerName
             if not gateway_name and rri.controllerUid:
-                all_gateways = gateway_helper.get_all_gateways(params)
-                for gateway in all_gateways:
-                    if gateway.controllerUid == rri.controllerUid:
-                        gateway_name = gateway.controllerName
-                        break
+                def _normalize_uid(uid):
+                    if uid is None:
+                        return None
+                    if isinstance(uid, (bytes, bytearray)):
+                        return utils.base64_url_encode(uid)
+                    return str(uid)
+
+                target_uid = _normalize_uid(rri.controllerUid)
+                try:
+                    all_gateways = gateway_helper.get_all_gateways(params) or []
+                except Exception:
+                    all_gateways = []
+
+                matched = next((g for g in all_gateways
+                               if _normalize_uid(getattr(g, 'controllerUid', None)) == target_uid), None)
+                if matched:
+                    gateway_name = getattr(matched, 'controllerName', None)
+                    logging.debug(f"Resolved gateway name from controllerUid {target_uid} -> {gateway_name}")
+
             gateway_name = gateway_name if gateway_name else '-'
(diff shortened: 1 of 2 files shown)


keepercommander/commands/discoveryrotation.py
```diff
@@ -3265,7 +3265,28 @@ def execute(self, params, **kwargs):
...
+                        return utils.base64_url_encode(uid)
+                    return str(uid)
+
+                target_uid = _normalize_uid(rri.controllerUid)
+                try:
+                    all_gateways = gateway_helper.get_all_gateways(params) or []
+                except Exception:
+                    all_gateways = []
+
+                matched = next((g for g in all_gateways
```
adeshmukh-ks, line 3283: is it possible that none and none would equate here and give false positive

keepercommander/commands/discoveryrotation.py
```diff
@@ -3265,7 +3265,28 @@ def execute(self, params, **kwargs):
...
+            gateway_name = rri.controllerName
+            if not gateway_name and rri.controllerUid:
+                def _normalize_uid(uid):
+                    if uid is None:
+                        return None
+                    if isinstance(uid, (bytes, bytearray)):
+                        return utils.base64_url_encode(uid)
+                    return str(uid)
+
+                target_uid = _normalize_uid(rri.controllerUid)
```
sali-ks, line 3277: The nested _normalize_uid() function handles bytes/string conversion, but the logic assumes all gateway objects have consistent controllerUid types. If mixed formats exist in the list, matching may fail unpredictably. Add explicit validation or normalize all UIDs at the point of retrieval.

keepercommander/commands/discoveryrotation.py
```diff
@@ -3265,7 +3265,28 @@ def execute(self, params, **kwargs):
...
+                try:
+                    all_gateways = gateway_helper.get_all_gateways(params) or []
+                except Exception:
+                    all_gateways = []
```
sali-ks: The except Exception: clause silently swallows all exceptions from gateway_helper.get_all_gateways(), including unexpected errors that should surface. Replace with specific exception types or add minimal logging to aid debugging.

keepercommander/commands/discoveryrotation.py
sshrushanth-ks: Yes, added a check to prevent the false positive

sali-ks (approved):

PR #2374, opened 2026-09-17
Update keeper-dag to support multple DATA edges with path notation.

jwalstra-keeper:
The prior keeper-dag would deactivate similar edges.
Added a toggle to prevent this if the edge had different paths.

In the PAM graph, resources have multiple DATA edges. 
The prior would disable them. 
The change now does not disable them.

The changes in Commander are for the debug commands in order to show information properly.


erinlewis-keeper (approved):
LGTM

PR #2378, opened 2026-09-18
Tailscale Funnel as a new Service Mode tunnel option

mnaqvi-ks:
Adds Tailscale Funnel as a new tunnel option for Commander Service Mode, alongside the existing Ngrok and Cloudflare integrations. Users can now expose the Service Mode REST API over a public HTTPS URL using their own Tailscale account, with no changes to how Ngrok/Cloudflare already work.

```bash
# Plain personal auth key
keeper service-create -p 8090 -ts '<tailscale-auth-key>' -rm foreground -q y -f yaml -c "ls"

# OAuth-issued auth key (needs the ACL tag it's scoped to)
keeper service-create -p 8090 -ts '<oauth-client-secret>' -tst 'tag:commander-service' -rm foreground -q y -f yaml -c "ls"
```

## Features

- **New tunnel provider** — `-ts, --tailscale <auth-key>` / `-tst, --tailscale_advertise_tags <tags>`, following the same single-flag convention as `-ng`/`-cf`. Mutually exclusive with Ngrok/Cloudflare.
- **CLI detection & auto-install** — detects a missing Tailscale CLI and, with explicit user confirmation, installs it automatically (Homebrew on macOS, official script on Linux, MSI with UAC elevation on Windows). Falls back to manual guidance if declined or if it fails.
- **Daemon detection & auto-start** — same consent-gated flow if `tailscaled` isn't running.
- **Authentication** — supports both plain personal auth keys and OAuth-client-issued keys. Used only for `tailscale up`; never touches the Service Mode API Key, which continues to gate every API request unchanged.
- **Full Funnel lifecycle** — start, live status, and reset wired into `service-create`/`service-start`/`service-status`/`service-stop`. Startup failures roll back cleanly (including already-started Ngrok/Cloudflare processes).
- **Public URL retrieval** — real HTTPS URL pulled from the node's Tailscale hostname; printed on create, persisted to config, reported via `service-status`, and included in Docker vault-record metadata (`-ur`).
- **Security** — auth key passed via a short-lived, owner-only temp file (`--auth-key=file:...`) instead of a raw CLI argument, avoiding exposure via `ps` to other local users. Never logged.

## Design notes

- `service-stop` only resets the Funnel exposure — it doesn't log the machine out of Tailscale or stop the daemon, since `tailscaled` is system-wide and may be used for other things on the same machine.
- CLI flag naming (`-ts <key>` as a single flag) intentionally matches the existing `-ng`/`-cf` convention rather than a separate enable+key pair.

## Files changed

| File | Purpose |
|---|---|
| `keepercommander/service/config/tailscale_config.py` *(new)* | Orchestrates CLI/daemon checks, authentication, Funnel start, and URL retrieval — the Tailscale equivalent of `ngrok_config.py`/`cloudflare_config.py` |
| `keepercommander/service/util/tunneling.py` | Adds all low-level `tailscale` CLI wrapper functions (install, daemon start, auth, Funnel start/stop/status, URL lookup) |
| `keepercommander/service/core/service_manager.py` | Wires Tailscale into `start_service`/`stop_service`/`get_status`, including rollback-on-failure and interrupt handling |
| `keepercommander/service/core/process_info.py` | Adds `tailscale_enabled`/`tailscale_port` lifecycle tracking (no PID — Tailscale has no Commander-owned subprocess) |
| `keepercommander/service/core/globals.py` | Transient, same-process global for deferring vault-metadata write-back until the real Funnel URL is known |
| `keepercommander/service/commands/create_service.py` | Adds `-ts`/`-tst` CLI flags and config schema fields |
| `keepercommander/service/commands/service_config_handlers.py` | Interactive and streamlined config prompts/assembly for Tailscale, extending the existing mutual-exclusion chain |
| `keepercommander/service/commands/integrations/vault_metadata.py` | Extracts the shared `get_service_url()` helper; fixes Docker metadata write-back timing so it waits for the real URL |
| `keepercommander/service/config/config_validation.py` | Adds `validate_tailscale_auth_key` |
| `keepercommander/service/config/models.py` / `service_config.py` | Adds `tailscale`, `tailscale_auth_key`, `tailscale_advertise_tags`, `tailscale_public_url` config fields, defaults, and backward-compat migration |
| `keepercommander/resources/service_config.ini` | Adds interactive prompt text for Tailscale |
| `keepercommander/service/README.md` | Documents the new flags, examples, and the `--force-reauth`/single-use-key note |
| `unit-tests/service/test_create_service.py` | Tests for the new CLI flag behavior, mutual exclusion, and streamlined config assembly |

## Compatibility

No changes to `ngrok_config.py` or `cloudflare_config.py`. All changes to shared files are additive, confirmed via diff review. Full existing test suite (279 tests) passes unchanged.


keepercommander/service/core/service_manager.py
```diff
@@ -170,14 +240,15 @@ def filter(self, record):
 
                     logger.debug(f"Service subprocess logs available at: {log_file}")
                     print(f"Commander Service started with PID: {process.pid}")
-                    ProcessInfo.save(process.pid, is_running, ngrok_pid, cloudflare_pid)
+                    ProcessInfo.save(process.pid, is_running, ngrok_pid, cloudflare_pid, tailscale_enabled=tailscale_enabled, tailscale_port=tailscale_port)
 
                 except Exception as e:
                     logger.error(f"Failed to start service subprocess: {e}")
                     raise
```
pkamble-ks, line 249: Tailscale Funnel is started before the Commander service subprocess is successfully created. If `spawn_detached_process()` or `ProcessInfo.save()` fails here, the exception path does not stop the Tailscale Funnel that was already configured at lines **118-120**. This can leave the Funnel publicly exposed while the Commander service is not running. Can we add Tailscale cleanup to this rollback path?

keepercommander/service/core/service_manager.py
pkamble-ks, line 592: When the Commander process is no longer running, this clears `ProcessInfo` but leaves the Tailscale Funnel configuration active. Since Funnel is managed by `tailscaled` and not tied to the Commander process lifecycle, an unexpected Commander crash/SIGKILL can leave the public Funnel endpoint enabled. Should we reconcile/disable the Commander-managed Funnel when detecting that the service process is gone?

keepercommander/service/util/tunneling.py
```diff
@@ -429,4 +429,454 @@ def generate_cloudflare_url(port, tunnel_token, custom_domain, run_mode):
...
+            try:
+                os.unlink(key_file_path)
+            except OSError:
+                pass
+
+
+# Tailscale Funnel only accepts one of these as the external-facing port;
+# the local target port (the Commander service port) is unrestricted and
+# separate. 443 is the default so the public URL needs no port suffix.
+TAILSCALE_FUNNEL_ALLOWED_PORTS = (443, 8443, 10000)
```
pkamble-ks, line 765: `TAILSCALE_FUNNEL_ALLOWED_PORTS` defines the supported external Funnel ports, but `start_tailscale_funnel()` doesn't validate `funnel_port` against it. Should we validate the port here before invoking the CLI, or remove the unused constant?

mnaqvi-ks: Added validation using the existing constant instead of removing it. The parameter is retained for future extensibility and keeps the signature consistent with `get_tailscale_funnel_url` and `stop_tailscale_funnel`, so validating it now prevents invalid values from reaching the Tailscale CLI and failing there.

keepercommander/service/core/service_manager.py
mnaqvi-ks: fixed, the outer exception handler now stops the Funnel if it was already up before the subprocess/Flask app failed to start.

keepercommander/service/core/service_manager.py
mnaqvi-ks: Fixed — get_status() now stops the Funnel too when it detects the process is gone, not just on an explicit service-stop

keepercommander/service/core/logs/tailscale_subprocess.log
mbachhav-ks, line 1: Do not commit log file and log folder

keepercommander/service/config/tailscale_config.py
```diff
@@ -0,0 +1,152 @@
...
+        Generic check -> guidance -> prompt -> attempt -> reverify flow, shared
+        by the CLI-install and daemon-start checks below. Raises ValidationError
+        if the user declines or the automatic attempt doesn't fix the check.
+        """
+        if check_fn():
+            return
+
+        guidance = guidance_fn()
+        logger.error(guidance)
+        print(guidance)
```
mbachhav-ks, line 64: Can you explain why guidance is printed?

mbachhav-ks (changes requested):

keepercommander/service/config/tailscale_config.py
mnaqvi-ks: This flow handles two checks: the Tailscale CLI being missing and the daemon not running. In either case, we show the manual fix before offering auto-fix, so the user always knows what to do if they decline or auto-fix fails
Guidance here printed for the manual fix

keepercommander/service/core/logs/tailscale_subprocess.log
mnaqvi-ks: Sorry for that 
fixed

